PR #4, opened 2018-04-06
Add release tooling to common repo

agjohnson:
This is a fork of the .org release tooling.


tasks.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Tag release of Read the Docs.
+
+    Do this after prepare task and manual cleanup/commit
+    """
+    # Ensure we're on the master branch first
+    git_rev_parse = ctx.run('git rev-parse --abbrev-ref HEAD', hide=True)
+    current_branch = git_rev_parse.stdout.strip()
+    if current_branch != 'master':
+        print('You must be on master branch!')
```
humitos: What about adding something like

```python
input("""
Hey! Did you already:
1. ran `invoke prepare`
2. organize CHANGELOG.rst removing invalid PRs
3. cleanup commits
4. etc...
? [y/N]""")
```

to avoid mistakes and remind what are the proper steps to be followed.

agjohnson: This would make a good enhancement. I think we can automate the checks though:

* Check setup.cfg for a matching version to the one you're releasing
* Check the changelog for a version string matching the one you're releasing
* Check that setupcfg and changelog changes are commited

PR #6, opened 2018-04-17
Ignore rtd_tests folder

stsewd:
The current patterns don't ignore all the folder

https://github.com/PyCQA/prospector/blob/master/prospector/profiles/profiles/no_test_warnings.yaml

### DIFF
--- a/prospector.yml
+++ b/prospector.yml
@@ -11,6 +11,7 @@ ignore-paths:
   - docs/
 
 ignore-patterns:
+  - rtd_tests/
   - /migrations/
   - local_settings.py
 


stsewd:
Also, should I sent the patch to prospector? or is very not common to have `*_tests/` folder?

stsewd:
@humitos because of https://github.com/rtfd/readthedocs.org/pull/3956, there are files only required by tests https://github.com/rtfd/readthedocs.org/tree/master/readthedocs/rtd_tests that are being checked by the linters. So, these files have some dependencies that are not required on the linting requirements. And prospector is configured to don't check tests files.

humitos:
I suppose we _do_ want to run prospector in our `rtd_tests` folder.

We have the default behavior (ignore) disabled:

https://prospector.landscape.io//en/master/profiles.html#test-warnings

Also, we have a simpler checks under the `prospector-more.yml` which is not so aggressive and there it's excluded: https://github.com/rtfd/common/blob/master/prospector-more.yml#L31

stsewd:
@humitos I don't understand this

> We have the default behavior (ignore) disabled:

https://github.com/rtfd/common/blob/aa3aa60c868295897dd0b141310b5e1853aa7d4f/prospector.yml#L3

Here is explicitly ignoring linting files on tests

ericholscher:
We do all sorts of wacky stuff in our tests, so I'm +1 on skipping linting them for now. We do lots of `import *` and monkeypatching and other stuff that will blow up.

PR #9, opened 2018-06-08
Automate repo label syncing

agjohnson:
Automates all the github label pieces so that our labels are the same across
any of our repos. New labels are from
https://github.com/rtfd/readthedocs.org/pull/4180

Also is the foundation for better autochangelog, using pygithub directly

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -5,6 +5,7 @@ from __future__ import division, print_function, unicode_literals
 
 import os
 import textwrap
+from collections import namedtuple
 
 from dateutil.parser import parse
 from future.moves.configparser import RawConfigParser, NoSectionError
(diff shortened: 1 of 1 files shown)


agjohnson:
Already did a pass on rtfd/redirect, went smoothly. Also tested against a live repo of mine really quick.

tasks.py
```diff
@@ -140,3 +155,142 @@ def get_config(config=None):
...
+        return False
+    for label in labels:
+        found = None
+
+        for existing_label in existing_labels:
+            if label.name == existing_label.name:
+                found = existing_label
+                break
+        if not found:
+            for existing_label in existing_labels:
```
humitos: I think you can simplify this by doing `existing_label.name in ([label.name] + label.transpose)` instead of having two loops.

agjohnson: The reason it is this way is to provide precedence to existing labels before it tries any transposed labels

PR #12, opened 2018-07-05
Prospector update config

humitos:
Add minor changes and comment explaining how to run it properly with similar configurations as we have in TravisCI.

### DIFF
--- a/pre-commit-config.yaml
+++ b/pre-commit-config.yaml
@@ -71,4 +71,11 @@ repos:
       # https://github.com/pre-commit/pre-commit/issues/178
       language: system
       files: '\.py$'
-      args: ['--profile=prospector']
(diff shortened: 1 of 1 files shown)


humitos:
The root problem is that we can't run `prospector` command from inside the venv created by pre-commit (because `readthedocs` and `readthedocsinc` are not found since we add them manually to PYTHONPATH and I didn't find a way to do the same inside the pre-commit venv).

So, the `language: system` line tells pre-commit to run the `prospector` command from the system directly (outside the venv) and for the `prospector` command to work properly we need to install the `lint` dependencies in the `venv` of the project itself.

That's is what I tried to explain in the comment added in this PR. Am I clear?

The other change in this PR is the exclusion of the `migrations` files.

agjohnson:
Ah gotcha. Yup, makes sense!

PR #13, opened 2018-07-05
Ignore missing docstrings in public module/class/method

humitos:
We don't want to fail the linting when there is no docstring in some places.

Related: https://github.com/rtfd/readthedocs.org/pull/4303#issuecomment-402814388

### DIFF
--- a/prospector.yml
+++ b/prospector.yml
@@ -22,6 +22,7 @@ pep8:
 
 pylint:
   options:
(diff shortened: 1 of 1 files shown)


agjohnson:
This fix isn't ideal, but I'm sure it's as close as we can get without pep257 being more configurable. 👍

PR #16, opened 2018-08-27
Fix cyclical changes and flake8 failure

agjohnson:
Flake8 just straight up fails and I don't know how to make it require a
later pydocstyle. The other two plugins just continually make cyclical
changes, mostly to __future__ imports, so i've removed them. I'm not
super familiar with the fallout though:

- What is adding __future__ in a bad format in the first place?
- Can we make python sort plugin use a not bad format?
- Can we set up order of plugins so at least changes are rewritten in
  the wrong order, creating cyclical changes?
- What is going to do python import sorting if this plugin is removed?

### DIFF
--- a/pre-commit-config.yaml
+++ b/pre-commit-config.yaml
@@ -2,10 +2,11 @@ exclude: '^$'
 fail_fast: false
 repos:
 
-- repo: https://github.com/asottile/add-trailing-comma
-  sha: v0.6.4
-  hooks:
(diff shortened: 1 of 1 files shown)


humitos:
This PR can be closed in favor of #18

agjohnson:
Closing this, but I don't think my original problem is solved by #18, we just need to put up with pre-commit being really confusing for now.

PR #17, opened 2018-08-30
Disable no-else-return

humitos:
We want this enabled, but it seems that it's buggy at the moment. We
exclude it for now.


stsewd:
But the reported files really have a return else.

humitos:
Closing as invalid.

I fixed my PR at https://github.com/rtfd/readthedocs.org/pull/4318/commits/3087f5bed8f1079b72b7b09e86397ddef0b74a3c by fixing all the places where this check failed.

PR #18, opened 2018-09-05
Improves for pre-commit configs

humitos:
Updated all the plugins and disabled the ones that were failing (`flake8`).

At the moment, we can't get to a status where `pre-commit` works exactly as we want. In other words, the flow we want is:

1. run `pre-commit`
1. some plugins fail but make their changes
1. some other plugins fail and changes are manually done
1. those changes are staged
1. run `pre-commit`
1. all green

Currently, this is not possible for different reasons:

1. we want to sort imports by using `isort` since it's the best one. Although, `yapf` has no option to disable touching the imports. Because of this reason, we need to run first `yapf` (will touch the import and so fail) and _after that_ call `isort` (will touch the imports and fail).
1. for multi lines that do not end on a comma, we want to add the trailing comma. To do this, we use `add-trailing-comma` plugin which is great. Although, we don't want this plugin to touch our imports but there is no option to disable it. Then, after this plugin is ran, `yapf` will also try to do something with the imports, and because of this, it will fail

```diff
# add-trailing-comma
from __future__ import (
-    absolute_import, division, print_function, unicode_literals)
+    absolute_import, division, print_function, unicode_literals,
+)
```

```diff
# yapf
 from __future__ import (
-    absolute_import, division, print_function, unicode_literals)
+    absolute_import,
+    division,
+    print_function,
+    unicode_literals,
+)
```

I configured `isort` to behaves as `yapf` for now to try to avoid this kind of issues. Although, I think the preferred way to sort the import is 

```python
from __future__ import (
    absolute_import, division, print_function, unicode_literals)
```

Anyway, once they implement these settings we need, changing this is really easy to make all of our code to follow the new standard.

> NOTE: the order of the calls of these plugins (`yapf`, `isort` and `add-trailing-comma`) is important. Otherwise, we will end up with a style that we don't want.

Related #16

### DIFF
--- a/isort.cfg
+++ b/isort.cfg
@@ -1,7 +1,23 @@
 [settings]
 line_length=80
-indent='    '
-multi_line_output=4
+indent=4
+
+# TODO: I think we want ``4 (Hanging Grid)`` but given the problems we
+# are having with yapf and add-trailing-comma; we are force to use 3.
+multi_line_output=3
+
+# TODO: yapf uses 1 line. Changing this setting will interfers with
+# yapf that doesn't have a way to set this value. I prefer 2.
+lines_after_imports=1
+
+combine_as_imports=False
+force_adds=False
+combine_star=False
+include_trailing_comma=True
+use_parentheses=True
+from_first=False
+force_alphabetical_sort=False
+
 default_section=THIRDPARTY
 known_first_party=readthedocs,readthedocsinc
 known_third_party=mock,builtins
--- a/pre-commit-config.yaml
+++ b/pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
       args: ['--in-place', '--remove-all-unused-imports', '--remove-unused-variable']
 
 - repo: git@github.com:pre-commit/mirrors-yapf.git
-  rev: v0.22.0
+  rev: v0.23.0
   hooks:
     - id: yapf
       exclude: 'migrations|settings|scripts'
--- a/pre-commit-config.yaml
+++ b/pre-commit-config.yaml
@@ -51,17 +51,19 @@ repos:
       'flake8-coding',
       'flake8-commas',
       'flake8-comprehensions',
-      'flake8-debugger',
       'flake8-deprecated',
       'flake8-docstrings',
       'flake8-meiqia',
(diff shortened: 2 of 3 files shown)


humitos:
I'd say that this is the best we can have for now and I'd like to merge this and keep improving it in the future when more settings on these plugins come up.

I've been testing this with `readthedocs/projects/models.py` and it works pretty good. Although, even after finishing the styling, our `pre-commit` keeps failing (because of the reasons that I mentioned before)

On the other hand, I also tested this with `readthedocs/builds/models.py` and after apply it twice, I got all green PASSED. My steps here were,

1. pre-commit run
1. git add
1. pre-commit run
1. git add
1. pre-commit run --> all GREEN

agjohnson:
The other option is that we change our expectations and just run yapf exclusively eventually. I think the two pass rule is going to confuse contributors until we automate things.

isort.cfg
```diff
@@ -1,7 +1,23 @@
...
-indent='    '
-multi_line_output=4
+indent=4
+
+# TODO: I think we want ``4 (Hanging Grid)`` but given the problems we
+# are having with yapf and add-trailing-comma; we are force to use 3.
+multi_line_output=3
+
+# TODO: yapf uses 1 line. Changing this setting will interfers with
+# yapf that doesn't have a way to set this value. I prefer 2.
```
agjohnson: Same. Good for now though

humitos:
With this config, even running pre-commit twice could mark two of the steps RED. In fact, it may happen it's impossible to get all of them GREEN because different plugins are competing for the same line.

Althouh, even when we could have 2 RED steps, since they are competing, no changes are generated to the files. Dumb example:

1. add-trailing-comma adds a comma -> RED
1. yapf removes the comma -> RED

these two "failed", but in the end, no files are modified and it's the state that we want.

I know, this is not the best but for now there is nothing we can do to fix this :/

agjohnson:
yeah, this was the case that I originally noted, it's really confusing when nothing changes. I feel like we eventually need to remove some of the plugins that are competing, if pre-commit has no protection for cyclical changes like this (which it probably should, like prospector aims to fix)

PR #20, opened 2018-10-30
Add pyupgrade

stsewd:
Pyupgrade https://github.com/asottile/pyupgrade/

This will help us to remove all the code for python2 support, I just tested it with one fiile, and looks good.

### DIFF
--- a/pre-commit-config.yaml
+++ b/pre-commit-config.yaml
@@ -85,3 +85,9 @@ repos:
       # project is installed because we are using ``language: system``
       # for this hook
       # additional_dependencies: []
+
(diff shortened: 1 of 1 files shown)


humitos:
Will it show errors or will it automatically change the code for us? Can we see an example of the result?

stsewd:
It will change the code for us, here is an example of what it does https://github.com/asottile/pyupgrade/#implemented-features. I tested it on the `projects/tasks.py` file, and it really does a good job.

humitos:
I'm closing this one since we already ran `pyupgrade` in org and com sites.

PR #21, opened 2018-11-06
Fix writing setup.cfg in binary mode.

ericholscher:
This was causing this error:

```
  File "/Users/eric/projects/readthedocs.org/tasks.py", line 65, in prepare
    config.write(configfile)
  File "/usr/local/Cellar/python/3.6.5_1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/configparser.py", line 919, in write
    self._sections[section].items(), d)
  File "/usr/local/Cellar/python/3.6.5_1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/configparser.py", line 923, in _write_section
    fp.write("[{}]\n".format(section_name))
TypeError: a bytes-like object is required, not 'str'
```

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -32,7 +32,7 @@ Label = namedtuple('Label', ['name', 'color', 'desc', 'transpose'])
 
 
 @task
-def prepare(ctx, version, since=None):
+def prepare(ctx, version, since=None, path=None):
     """
     Prepare the next release version by updating files.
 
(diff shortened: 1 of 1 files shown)


ericholscher:
Additional commit also allows passing a path explicitly to the `prepare` function, because it's broken on the .org currently. It tries to use `readthedocs.org/common/setup.cfg` because of the `REPO_PATH = os.path.dirname(__file__)`

tasks.py
```diff
@@ -32,7 +32,7 @@
 
 
 @task
-def prepare(ctx, version, since=None):
+def prepare(ctx, version, since=None, path=None):
```
humitos: You can just do `path=REPO_PATH` and no `if` is needed later.

PR #22, opened 2018-11-28
Update our linting config

humitos:
Many updates to fit our linting preferences.

### DIFF
--- a/isort.cfg
+++ b/isort.cfg
@@ -6,10 +6,7 @@ indent=4
 # are having with yapf and add-trailing-comma; we are force to use 3.
 multi_line_output=3
 
-# TODO: yapf uses 1 line. Changing this setting will interfers with
-# yapf that doesn't have a way to set this value. I prefer 2.
-lines_after_imports=1
-
+lines_after_imports=2
 combine_as_imports=False
 force_adds=False
 combine_star=False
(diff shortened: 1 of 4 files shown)


ericholscher:
I'd much prefer to just use `black`, which has no configuration options, and will stop us from having to update these configs over time. In particular, it is starting to become a standard for python libs, similar to gofmt, but mainly I just really want to not have to configure our linters.

humitos:
I never tried `black` but I saw that it's getting famous :)

I started using `yapf` because I had experience with it and I was able to match all the linting requirements that were needed.

@ericholscher `black` would be the replacement of what (`yapf`, `isort`, another or... all of them)?

stsewd:
@humitos black what I like form black is that it starts and ends with the same AST, for that same reason it can't do the same as isort (https://github.com/ambv/black/issues/333), but I've seen that people use black together with https://github.com/asottile/reorder_python_imports.

What I don't like about black is that uses double quotes, what I like is that we don't have to write big config files.

What I don't like about black is the 88 columns by default (it's configurable), what I like about black is how it does the line breaks and wrapping, it's very smart.

What I don't like... No, that's all p: +1 on using black

humitos:
OK. I will take some time in the following days to read about `black` and think if we should swap our current auto-linting to it.

I'm merging this PR since it doesn't interfers with the idea of using black or not.

PR #23, opened 2018-11-29
Initial configuration for pre-commit with black

humitos:
We change the usage of yapf plus a very customized group of external plugins to only rely on black (for auto-linting) + flake8 (warning checks) + isort (auto-sort import) + docformatter (auto-lint docstring)

This configuration is way more easy to mantain since there is not too many options to configure the auto linting.

The only configuration files are the ones for the warning checks that can be adapted/tunned over time if needed.

Related #22

### DIFF
--- a/flake8
+++ b/flake8
@@ -1,3 +1,36 @@
 [flake8]
-ignore = C818,C819,E125,D100,D101,D102,D103,D105,D106,D107,D200,D202,D211,D403,P101,FI15,FI16,FI12,FI11,FI17,FI50,FI53,FI54,MQ101,T000
-max-line-length = 80
+ignore =
+  C812,  # missing trailing comma
+  C818,
+  C819,
+  D100,
+  D101,
+  D102,
+  D103,
+  D105,
+  D106,
+  D107,
+  D200,
+  D202,
+  D211,
+  D403,
+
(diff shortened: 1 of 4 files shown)


humitos:
If you want to check how this configuration affects the code, you can checkout this PR in your .org code and run `pre-commit run --files` with the files you want.

One of the good news here is that running multiple times this command should produce always the same output and the different plugins shouldn't fight against them to modify one file and get modified again for the next plugin. So, no further confusions should be prompted with this flow.

ericholscher:
I'm a big +1. Arguing over syntax is one of my least favorite things, so I really like the idea of not having a configurable linter, and just using the style it outputs :)

agjohnson:
The problem isn't arguing over configurable linting. The work that @humitos did to configure yapf solved that. The problem is that our configuration is now sufficiently complex that each pre-commit step produces competing changes, resulting in consistent changes, but results in linting errors/changes to files every pass.

Black might solve this, or cutting down on our pre-commit plugins might solve this. I lean towards reducing plugins and keeping a configurable linter which is already tuned.

If black gives comparable output and avoids some of the more annoying parts of pep8 formatting, black is probably fine too though.

humitos:
Closing this as we are not going to use black for now.

PR #24, opened 2018-12-05
Update prospector configs

humitos:


ericholscher:
Any more details here? Not really sure how to review this.

humitos:
It's more a cleanup than other.

Basically, 

* it starts failing when using modular formatting on logging
* and fails when inheriting from `object`
* does not allow new line after function docstring (which matches with our yapf styling)

PR #25, opened 2018-12-11
Use default isort lines after import statements

agjohnson:
This is not ideal, but solves the issue of a cyclical lint fix when yapf
and isort disagree on the number of lines after imports.

### DIFF
--- a/isort.cfg
+++ b/isort.cfg
@@ -6,11 +6,7 @@ indent=4
 # are having with yapf and add-trailing-comma; we are force to use 3.
 multi_line_output=3
 
-# I dislike 1 line after imports, and PEP8 says 2, but yapf insists on 1 line.
-# Setting isort to 1 line here avoids a circular lint fix situation. Feature
-# request on yapf for this is:
-# https://github.com/google/yapf/issues/347
-lines_after_imports=1
+lines_after_imports=-1
 combine_as_imports=False
 force_adds=False
 combine_star=False
--- a/isort.cfg
+++ b/isort.cfg
@@ -6,7 +6,11 @@ indent=4
 # are having with yapf and add-trailing-comma; we are force to use 3.
 multi_line_output=3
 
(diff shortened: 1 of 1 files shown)


agjohnson:
This might also take a config change to flake8. The other option is to make this change in yapf instead. Ultimately, it only really affects us when using pre-commit to ensure lint fixes, as this is where circular problems will be frustrating

agjohnson:
Oh, and super annoying, the 1 vs 2 line thing isn't always a problem. I'm a bit confused now and don't know if this PR fixes the issue at all.

stsewd:
Here is the issue in yapf https://github.com/google/yapf/issues/347

humitos:
This is one could be another approach that help us here: https://github.com/google/yapf/issues/429

humitos:
I suppose this is the chunk of code you refers to

https://github.com/timothycrosley/isort/blob/27515dbf9ba2137bd62f766d930273232338d953/isort/isort.py#L591-L597

> This forces the default behavior that unless the following line is a function or class definition, lines after imports will be 1 line, otherwise 2 -- instead of always forcing 2.

My understanding is that it works as you say here.

humitos:
I think if this change does not produce a `propector` to fail, we are OK. Otherwise, we should check which propector plugin fail and disable that check or similar.

agjohnson:
Roger. How can we confirm that? I don't believe I ran into any problems with prospector while linting the code changes i had. I could see maybe somewhere later we hit a problem with prospector though. Should we just try this fix?

humitos:
Yeah. Let's try this fix.

You can test local prospector with pre-commit first, as well.

arareko:
A knob `BLANK_LINES_BETWEEN_TOP_LEVEL_IMPORTS_AND_VARIABLES` will be available in version `0.31.0` of YAPF

PR #29, opened 2019-01-20
Mark recommonmark to not be upgraded

humitos:
Newer `recommonmark` is not compatible with our code.

See https://github.com/rtfd/readthedocs.org/pull/5134


ericholscher:
We should probably just fix our integration. I will do a PR for that.

PR #41, opened 2020-01-13
Use devpi for PyPI cache

humitos:
Related to https://github.com/readthedocs/readthedocs.org/pull/6340/


humitos:
This PR needs to be adapted with the new changes for shared common files.

PR #45, opened 2020-01-22
Add start/stop commands for docker-compose

stsewd:


humitos:
I'm not sure how you use `inv docker.`.

I'm just doing always `inv docker.up` and it _only creates_ the container if it does not exist already (first time). Then, just Ctrl-C makes the containers to stop --and do not delete them.

Next time I run `inv docker.up` it just call `start` behind the scenes.

So, to me adding `start` and `stop` won't have any effect.

stsewd:
@humitos I don't want to force a build of the images sometimes, like when changing branches. Also, `start` starts all services in the background, so I can use the current shell after running docker.start.

humitos:
> @humitos I don't want to force a build of the images sometimes, like when changing branches.

if you change branches and there is a re-build, there is something wrong there. That should not happen at all.

stsewd:
> if you change branches and there is a re-build, there is something wrong there. That should not happen at all.

If you change a branch that has changes in the docker-compose files.

humitos:
> If you change a branch that has changes in the docker-compose files.

In that case you need to do `down` and then `up` for those changes to be read.

stsewd:
> In that case you need to do down and then up for those changes to be read.

My point is to not rebuild, also more benefits from https://github.com/readthedocs/common/pull/45#issuecomment-579872471 and having more control over what process start.

I'm using `inv docker.compose command` now anyway

PR #48, opened 2020-01-25
Reduce search container ram to 128Mb

humitos:
There is no need to use 512Mb.

Container memory is reduced from ~800 to ~400.


humitos:
Good. We can increased it again if we feel that it's required. I'm happy to not consume a lot of memory for nothing at the moment.

PR #49, opened 2020-02-11
`compose` task to pass the command directly to docker-compose

humitos:
This can be used as `inv docker.compose 'stop nginx'` for example.

It's just a helper to call docker-compose without specifying all the
.yaml files by hand.

### DIFF
--- a/dockerfiles/tasks.py
+++ b/dockerfiles/tasks.py
@@ -11,6 +11,11 @@ def build(c):
     c.run(f'{DOCKER_COMPOSE_COMMAND} build --no-cache', pty=True)
 
 @task
+def compose(c, command):
+    """Pass the command to docker-compose directly."""
+    c.run(f'{DOCKER_COMPOSE_COMMAND} {command}', pty=True)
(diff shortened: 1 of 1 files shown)


humitos:
@stsewd do you know a way to make this possible?:

`inv docker.compose stop nginx` instead `inv docker.compose 'stop nginx'` (note the single quotes.

humitos:
basically, what I want is to pass "anything after the `docker.compose` as the `command` argument"

stsewd:
oh, didn't realize that, maybe declaring the `command` as commands? (or forcing it to be multi-arg)

stsewd:
From here https://github.com/pyinvoke/invoke/issues/378 looks like isn't possible at the moment... Some people are using https://magicinvoke.readthedocs.io/en/latest/examples/args-kwargs/README.html for that

humitos:
OK. I will merge this as is then.

I don't think we want to change from invoke to magicinvoke just for this. Looks interesting, though.

PR #50, opened 2020-02-19
Add more options to docker tasks

agjohnson:
I found these two operations cumbersome. I don't need all the build
images locally, and in testing changes to build the image, I wanted
caching (I was stuck redownloading dependencies over cafe wifi). Not
caching is sometimes preferable, so we should at least allow the option.
Default is false on both.

I updated the list of images, not sure if it makes sense to also include
the testing image by default though.


humitos:
You may want to update the docs at https://docs--6681.org.readthedocs.build/en/6681/development/standards.html#working-with-docker-compose

agjohnson:
I think that sort of documentation should be *in* the tasks.py, so the documentation on what tasks do and what the options mean is right where the user is using the script, not in a separate doc. Documentation is a great place to describe the flow of commands I need to run, but I'd run `inv -h` to get information on using the scripts in more detail.

It's something that could potentially be duplicated in the docs, like we do in ops via autodoc on the tasks.py file, but that doesn't seem super important.

PR #51, opened 2020-02-19
Add ext-theme and path configuration

agjohnson:
I would prefer configurable paths here. If you set environment variables
RTDDEV_PATH_*, this will bubble up to docker compose.

There will be two related PRs that enable theme usage: one for community
and one for corporate.

### DIFF
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -29,7 +29,8 @@ services:
     volumes:
       - ${PWD}/common/dockerfiles/entrypoints/common.sh:/usr/src/app/docker/common.sh
       - ${PWD}/common/dockerfiles/entrypoints/proxito.sh:/usr/src/app/docker/proxito.sh
-      - ${PWD}/../readthedocs-ext:/usr/src/app/checkouts/readthedocs-ext
+      - ${PWD}/${RTDDEV_PATH_EXT:-../readthedocs-ext}:/usr/src/app/checkouts/readthedocs-ext
+      - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../readthedocs-ext-theme}:/usr/src/app/checkouts/ext-theme
     links:
       - storage
       - database
@@ -51,7 +52,8 @@ services:
       - ${PWD}/common/dockerfiles/scripts/createsuperuser.py:/usr/src/app/docker/createsuperuser.py
       - ${PWD}/common/dockerfiles/entrypoints/common.sh:/usr/src/app/docker/common.sh
       - ${PWD}/common/dockerfiles/entrypoints/web.sh:/usr/src/app/docker/web.sh
-      - ${PWD}/../readthedocs-ext:/usr/src/app/checkouts/readthedocs-ext
+      - ${PWD}/${RTDDEV_PATH_EXT:-../readthedocs-ext}:/usr/src/app/checkouts/readthedocs-ext
+      - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../readthedocs-ext-theme}:/usr/src/app/checkouts/ext-theme
     links:
       - storage
       - database
@@ -61,6 +63,7 @@ services:
     environment:
       - INIT=${INIT:-}
       - DOCKER_NO_RELOAD
+      - RTD_EXT_THEME_ENABLED
     stdin_open: true
     tty: true
     networks:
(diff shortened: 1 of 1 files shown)


dockerfiles/docker-compose.yml
```diff
@@ -29,7 +29,8 @@ services:
     volumes:
       - ${PWD}/common/dockerfiles/entrypoints/common.sh:/usr/src/app/docker/common.sh
       - ${PWD}/common/dockerfiles/entrypoints/proxito.sh:/usr/src/app/docker/proxito.sh
-      - ${PWD}/../readthedocs-ext:/usr/src/app/checkouts/readthedocs-ext
+      - ${PWD}/${RTDDEV_PATH_EXT}:/usr/src/app/checkouts/readthedocs-ext
+      - ${PWD}/${RTDDEV_PATH_EXT_THEME}:/usr/src/app/checkouts/readthedocs-ext-theme
```
stsewd: You could also use defaults `${RTDDEV_PATH_EXT:-../readthedocs-ext}`

https://docs.docker.com/compose/compose-file/#variable-substitution

agjohnson: Thanks! I was looking for something like this and wasn't finding it in the docs. That's a much better solution

dockerfiles/docker-compose.yml
```diff
@@ -29,7 +29,8 @@ services:
     volumes:
       - ${PWD}/common/dockerfiles/entrypoints/common.sh:/usr/src/app/docker/common.sh
       - ${PWD}/common/dockerfiles/entrypoints/proxito.sh:/usr/src/app/docker/proxito.sh
-      - ${PWD}/../readthedocs-ext:/usr/src/app/checkouts/readthedocs-ext
+      - ${PWD}/${RTDDEV_PATH_EXT:-../readthedocs-ext}:/usr/src/app/checkouts/readthedocs-ext
+      - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../readthedocs-ext-theme}:/usr/src/app/checkouts/readthedocs-ext-theme
```
humitos: What repository is `readthedocs-ext-theme`? I'd say the default should be the same name of the repository (I suppose it's https://github.com/readthedocs/ext-theme, which the default would be `ext-theme` in that case)

agjohnson:
This has been sitting for a while, so just going to request re-review here. Is env var naming okay here?

dockerfiles/docker-compose.yml
```diff
@@ -61,6 +63,7 @@ services:
     environment:
       - INIT=${INIT:-}
       - DOCKER_NO_RELOAD
+      - RTD_EXT_THEME_ENABLED
```
humitos, line 66: Where this value come from? `inv docker.up` command with a flag?

dockerfiles/docker-compose.yml
```diff
@@ -72,7 +75,8 @@ services:
       - ${PWD}/common/dockerfiles/entrypoints/common.sh:/usr/src/app/docker/common.sh
       - ${PWD}/common/dockerfiles/entrypoints/celery.sh:/usr/src/app/docker/celery.sh
       - ${PWD}/common/dockerfiles/scripts/wait_for_search.py:/usr/src/app/docker/scripts/wait_for_search.py
-      - ${PWD}/../readthedocs-ext:/usr/src/app/checkouts/readthedocs-ext
+      - ${PWD}/${RTDDEV_PATH_EXT:-../readthedocs-ext}:/usr/src/app/checkouts/readthedocs-ext
+      - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../ext-theme}:/usr/src/app/checkouts/ext-theme
```
humitos, line 79: I'm OK using `RTDDEV_PATH_EXT` for `readthedocs-ext` path here. However, I don't think that `ext-theme` is required in `celery`, `build` and `proxito`. I'd say that it may be required only in `web` probably, right?

humitos (commented):
Looks good. I think we are adding the theme in containers where it's not required but I don't have the full context to be able to test this.

dockerfiles/docker-compose.yml
agjohnson: Yes, just from the task, but you could also specify in `.envrc` as a side effect

dockerfiles/docker-compose.yml
agjohnson: I assume it isn't needed at all, but is there a strong reason not to include it in these containers?

stsewd (approved):
I think we only need this in the web service - but no strong opinion.

dockerfiles/docker-compose.yml
humitos: If it's not required, I won't include it. Keeping consistency with the rest is good to me and also adding things that "do nothing" just adds confusions in my experience; considering that our docker compose is overwritten in different places.

humitos:
Isn't this PR replaced by #73?

agjohnson:
No, #73 includes the commits here

dockerfiles/docker-compose.yml
agjohnson: I'd disagree, I think the larger issue is running instances that don't all have the same code. In production, we don't have instances that only have select repositories, all of the instances have the same pile of code. Developing with only a partial checkout seems like a way to miss side effects from repositories

humitos: Well, the same question is valid for production then: why are we installing things in the images that are not required? :smile:

humitos (approved):
I don't want to block this on https://github.com/readthedocs/common/pull/51/files#r507926876 but I think we shouldn't include code in the containers that it's not required.

dockerfiles/docker-compose.yml
agjohnson: There is a big argument for consistency. Making our images/docker images have special checkout requirements would be a recipe for introducing non-deterministic bugs around repository interactions.

On the other side of this argument, what is the downside of having additional repository definitions? That we have extraneous code lying around? That seems very minor in comparison to potentially spending time hunting bugs or introducing downtime with uncaught bugs.

PR #54, opened 2020-03-11
Create `envs` container for build environment cache

humitos:
Required by https://github.com/readthedocs/readthedocs.org/pull/6763


humitos:
It's suppose that it does not require any manual step. Once the common submodule is updated in the other repos this will be pulled automatically and this new command will create the bucket on next `inv docker.up`

PR #56, opened 2020-03-24
Reduce number of files watched by watchmedo

humitos:
Only watch `readthedocs/*.py` and `readthedocsinc/*.py`. This may help
a little with the CPU usage when running with reload enabled.

### DIFF
--- a/dockerfiles/entrypoints/build.sh
+++ b/dockerfiles/entrypoints/build.sh
@@ -16,6 +16,9 @@ else
   --ignore-patterns="*.pyo" \
   --ignore-patterns="*.pyc" \
   --ignore-patterns="*flycheck*.py" \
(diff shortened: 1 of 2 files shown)


humitos:
This PR won't be needed if we decide to use https://github.com/readthedocs/readthedocs.org/pull/6812

davidfischer:
Does `readthedocs/*.py` recurse subfolders?

humitos:
Yes. But the pattern was incorrect :/

You can test this out with,

```
$ watchmedo shell-command --patterns="./readthedocs/*.py" --recursive --command='echo "${watch_src_path}"' .
./readthedocs/builds/admin.py
./readthedocs/builds/storage/tsettes.py
```

PR #57, opened 2020-03-24
Rely on Django web app to reload Python modules via pool_restart

humitos:
This is a different approach to #56 that it's simpler and use Celery internals to reload all Python modules instead of killing and re-starting the whole process.

See https://github.com/readthedocs/readthedocs.org/pull/6812


humitos:
The required PR was closed. So, we can close this one for now as well.

PR #58, opened 2020-04-02
Consume from new 'build:' queues

humitos:
Make builders consume from the new queue names. See https://github.com/readthedocs/readthedocs.org/pull/6853


ericholscher:
This got fixed in our ops repo.

PR #60, opened 2020-04-28
Auto-reload fix

humitos:
This problem was introduced by myself at https://github.com/readthedocs/common/pull/56

It's not possible to use multiple args for `--patterns` and `--ignore-patterns` :(


humitos:
Yes. I tested that and editing a migration file does not trigger the auto-reload

PR #61, opened 2020-05-25
Use celery concurrency 1 and allow `--scale` containers

humitos:
This will give us similar parity with production, where there can't be 2 builds
building in the same server at the same time.

This can be called by,

```
inv docker.up --scale-build 3
```

### DIFF
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -1,8 +1,7 @@
 version: '3'
 
 volumes:
-  build-default-user-builds:
-  build-large-user-builds:
+  build-user-builds:
   storage_data:
   postgres_data:
   postgres_backups_data:
@@ -88,7 +87,7 @@ services:
       readthedocs:
     command: ["../../docker/celery.sh"]
 
-  build-default: &build
+  build:
     volumes:
       - ${PWD}/common/dockerfiles/entrypoints/common.sh:/usr/src/app/docker/common.sh
       - ${PWD}/common/dockerfiles/entrypoints/build.sh:/usr/src/app/docker/build.sh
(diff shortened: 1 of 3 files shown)


dockerfiles/docker-compose.yml
```diff
@@ -108,12 +109,35 @@ services:
...
       - DOCKER_NO_RELOAD
+      - BUILD_USER_BUILDS_VOLUME=build-default-user-builds
+      - CELERY_QUEUE=build:default
     stdin_open: true
     tty: true
     networks:
       readthedocs:
     command: ["../../docker/build.sh"]
 
+  build-large:
```
ericholscher, line 120: I feel like this is going to add even more docker overhead to our build setup. Is there a reason not to just have a single process that listens on both queues?

humitos: I'm not worried about having one builder per queue --In fact, now that you mentioned it, I probably want both builders to listen on both queues.

With this PR I want to achieve two different things:

- use single process builders, as we do in production
- keep the ability to build more than one build at a time: allow us to test things like concurrency limit, de-duplication, etc

However, if we are worried about spinning up another container, we could make this container "extra" builder to spin only if an argument is passed, similar to `search`.

humitos: Also, there is a path that I haven't explore yet, but `docker-compose up` supports scale:

```
    --scale SERVICE=NUM        Scale SERVICE to NUM instances. Overrides the
                               `scale` setting in the Compose file if present.
```

So, it may be better (if works) to have _only one single-process builder_ listening on both queues where we can control the number of instances with this `--scale` argument.

ericholscher: Yea, I think something like that is better. I'd really like to reduce the overhead of our docker setup, not add to it, since it's pretty heavy already.

humitos: I just tested `--scale` and it worked! I will try to adapt this PR to make allow its usage :100:

agjohnson: Yup, another vote for making this optional

PR #66, opened 2020-07-09
Ignore acl/sso/models.py because of a pylint-django bug

humitos:


stsewd:
Believe this is already fixed in master, and there are other valid options to make it work (`self`, `app.Model`, or the class itself)

stsewd:
https://github.com/readthedocs/readthedocs-corporate/pull/1002

PR #73, opened 2020-10-19
 Add docker and docker task support for using ext-theme

agjohnson:
This enables a few pieces needed for local development of the ext-theme
repository:

* Docker can use ext-theme without webpack, built assets are synced to
  storage
* Docker can use ext-theme without webpack, development server starts
  and assets are hot reloaded during development

The `docker.up` tasks was reworked to support the additional
configuration in an easier way. Some of the arguments are flipped,
because invoke automatically gives us command arguments like
`--[no-]search` if you specify `search=True`. The `docker.up` help is
currently:

    Usage: inv[oke] [--core-opts] docker.up [--options] [other tasks here ...]

    Docstring:
      Start all the docker containers for a Read the Docs instance

    Options:
      -e, --ext-theme-enabled   Enable new theme from ext-theme (default: False)
      -i, --init                Perform initialization steps (default: False)
      -r, --[no-]reload         Enable automatic process reloading (default: True)
      -s, --[no-]search         Start search container (default: True)
      -w, --webpack             Start webpack development server (default: False)

Requires #51

### DIFF
--- a/dockerfiles/docker-compose-webpack.yml
+++ b/dockerfiles/docker-compose-webpack.yml
@@ -7,17 +7,15 @@ services:
   webpack:
     image: node:12
     working_dir: /usr/src/app/checkouts/ext-theme
-    environment:
-      - NODE_ENV=development
     ports:
       - "10001:10001"
     volumes:
-      - ${PWD}/common/dockerfiles/entrypoints/common.sh:/usr/src/app/docker/common.sh
       - ${PWD}/common/dockerfiles/entrypoints/webpack.sh:/usr/src/app/docker/webpack.sh
       - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../readthedocs-ext-theme}:/usr/src/app/checkouts/ext-theme
     environment:
       - INIT=${INIT:-}
       - DOCKER_NO_RELOAD
+      - NODE_ENV=development
       - RTD_EXT_THEME_ENABLED
       - RTD_EXT_THEME_DEV_SERVER_ENABLED
     stdin_open: true
--- a/dockerfiles/docker-compose-webpack.yml
+++ b/dockerfiles/docker-compose-webpack.yml
@@ -11,7 +11,7 @@ services:
       - "10001:10001"
     volumes:
       - ${PWD}/common/dockerfiles/entrypoints/webpack.sh:/usr/src/app/docker/webpack.sh
-      - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../readthedocs-ext-theme}:/usr/src/app/checkouts/ext-theme
+      - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../ext-theme}:/usr/src/app/checkouts/ext-theme
     environment:
       - INIT=${INIT:-}
       - DOCKER_NO_RELOAD
--- a/dockerfiles/docker-compose-webpack.yml
+++ b/dockerfiles/docker-compose-webpack.yml
@@ -1,0 +1,27 @@
+# docker-compose-webpack.yml starts Webpack dev server for hot reload of built
+# static assets.
+version: '3'
+
+services:
+
+  webpack:
+    image: node:12
+    working_dir: /usr/src/app/checkouts/ext-theme
+    environment:
+      - NODE_ENV=development
+    ports:
+      - "10001:10001"
+    volumes:
+      - ${PWD}/common/dockerfiles/entrypoints/common.sh:/usr/src/app/docker/common.sh
+      - ${PWD}/common/dockerfiles/entrypoints/webpack.sh:/usr/src/app/docker/webpack.sh
+      - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../readthedocs-ext-theme}:/usr/src/app/checkouts/ext-theme
+    environment:
+      - INIT=${INIT:-}
+      - DOCKER_NO_RELOAD
+      - RTD_EXT_THEME_ENABLED
+      - RTD_EXT_THEME_DEV_SERVER_ENABLED
+    stdin_open: true
+    tty: true
+    networks:
+      readthedocs:
+    command: ["../../docker/webpack.sh"]
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -64,6 +64,7 @@ services:
       - INIT=${INIT:-}
       - DOCKER_NO_RELOAD
       - RTD_EXT_THEME_ENABLED
+      - RTD_EXT_THEME_DEV_SERVER_ENABLED
     stdin_open: true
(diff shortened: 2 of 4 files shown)


agjohnson:
Looks like my change to tasks.py clobbered a previous change I'll need to bring in

dockerfiles/docker-compose-webpack.yml
```diff
@@ -0,0 +1,27 @@
...
+    image: node:12
+    working_dir: /usr/src/app/checkouts/ext-theme
+    environment:
+      - NODE_ENV=development
+    ports:
+      - "10001:10001"
+    volumes:
+      - ${PWD}/common/dockerfiles/entrypoints/common.sh:/usr/src/app/docker/common.sh
+      - ${PWD}/common/dockerfiles/entrypoints/webpack.sh:/usr/src/app/docker/webpack.sh
+      - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../readthedocs-ext-theme}:/usr/src/app/checkouts/ext-theme
```
stsewd, line 17: Believe this is ext-theme now?

```suggestion
      - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../ext-theme}:/usr/src/app/checkouts/ext-theme
```

dockerfiles/docker-compose.yml
```diff
@@ -61,6 +63,8 @@ services:
     environment:
       - INIT=${INIT:-}
       - DOCKER_NO_RELOAD
+      - RTD_EXT_THEME_ENABLED
+      - RTD_EXT_THEME_DEV_SERVER_ENABLED
```
stsewd, line 67: We should put a default value here, if we don't pass this env var docker will raise a warning.

stsewd (approved):
I left some comments

dockerfiles/docker-compose-webpack.yml
```diff
@@ -0,0 +1,27 @@
...
+
+services:
+
+  webpack:
+    image: node:12
+    working_dir: /usr/src/app/checkouts/ext-theme
+    environment:
+      - NODE_ENV=development
+    ports:
+      - "10001:10001"
```
stsewd, line 13: Do we really need exposing this port int the host?

agjohnson: Yes, client requests for the hot reload files go directly to webpack dev server

dockerfiles/docker-compose.yml
agjohnson: I haven't noticed this at all, and do not have the env variables set. I was able to bring up the instance without webpack/theme -- that is, these env variables were not set

dockerfiles/tasks.py
```diff
@@ -26,21 +27,41 @@ def down(c, volumes=False):
...
         c.run(f'{DOCKER_COMPOSE_COMMAND} down', pty=True)
 
-@task
-def up(c, no_search=False, init=False, no_reload=False, scale_build=1):
+@task(help={
+    'search': 'Start search container (default: True)',
+    'init': 'Perform initialization steps (default: False)',
+    'reload': 'Enable automatic process reloading (default: True)',
+    'webpack': 'Start webpack development server (default: False)',
+    'ext-theme-enabled': 'Enable new theme from ext-theme (default: False)',
```
humitos, line 35: We can call it just `--ext-theme`

dockerfiles/docker-compose-webpack.yml
```diff
@@ -0,0 +1,27 @@
...
+# static assets.
+version: '3'
+
+services:
+
+  webpack:
+    image: node:12
+    working_dir: /usr/src/app/checkouts/ext-theme
+    environment:
+      - NODE_ENV=development
```
humitos, line 11: We have two `environment` keys here. This should be moved with the others.

dockerfiles/entrypoints/webpack.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/sh
+
```
humitos, line 2: We use a `common.sh` script to be ran in all the containers that it's usually called here. I'm not sure if it OK to add it here since this image won't have Python installed and may break when using `pip install whatever` in the common script.

If we are not calling it here, we should remove it from the `volumen:` in the `webpack` service.

humitos (commented):
I pull down this changes together with https://github.com/readthedocs/readthedocs.org/pull/7578/ and ran `inv docker.up --webpack --ext-theme-enabled` and I saw that everything built properly. However, I don't see the new theme when hitting http://community.dev.readthedocs.io/

Do I need to do something else to test this out?

I tried rebuilding the image `inv docker.build` and I got the problem you reported. How do I solve that?

```
Step 13/18 : COPY requirements/pip.txt pip.txt
ERROR: Service 'server' failed to build: COPY failed: stat /var/lib/docker/tmp/docker-builder563117358/requirements/pip.txt: no such file or directory
```

agjohnson:
@humitos it was a bad rebase, there was a more recent change that altered the path used that was missed. I can try rebasing this PR with #51 merged.

dockerfiles/entrypoints/webpack.sh
agjohnson: Yeah, this container is closer to Azurite -- it doesn't contain any application code, it's an external service. I'd agree we probably don't want to run common.sh here.

agjohnson:
This should be rebased now. I cleaned up some of the suggestions with f51a33d.

@humitos you also probably need the readthedocs/readthedocs.org@agj/ext-theme branch locally with the addtional settings.

dockerfiles/docker-compose-webpack.yml
humitos: it seems that you lost this with the rebase, which makes my local env to fail

humitos:
> @humitos you also probably need the readthedocs/readthedocs.org@agj/ext-theme branch locally with the addtional settings.

That branch seems it was already merged into master. I was able to build now (I had to change a path only), but I don't see the new templates when accessing the dashboard. What I'm missing?

humitos:
OK, it worked! I was calling `inv docker.build` without exporting `GITHUB_TOKEN` so -ext and ext-theme weren't installed which makes the feature to not work

humitos (approved):
Looks good! The only missing thing is to fix the path that santos commented and we lost in the rebase.

dockerfiles/docker-compose-webpack.yml
agjohnson: Oops, let's try this again!

dockerfiles/docker-compose.yml
agjohnson: I've held off on this because it requires more changes to rtd settings. If anyone notices any issues with warnings here, we can revisit, but I have not encountered any warnings about the missing variables.

PR #75, opened 2020-10-21
Add `running` flag to `docker.manage` task

agjohnson:
This allows for running things like `collectstatic` from a running
instance:

    inv docker.manage --running collectstatic

Instead of bringing up a new container, an existing container will be
used instead. This allows for the running `web` instance settings to be
reused.

### DIFF
--- a/dockerfiles/tasks.py
+++ b/dockerfiles/tasks.py
@@ -72,9 +72,12 @@ def shell(c, running=False, container='web'):
         c.run(f'{DOCKER_COMPOSE_COMMAND} run --rm {container} /bin/bash', pty=True)
 
 @task
-def manage(c, command):
+def manage(c, command, running=False):
     """Run manage.py with a specific command."""
-    c.run(f'{DOCKER_COMPOSE_COMMAND} run --rm web python3 manage.py {command}', pty=True)
+    subcmd = 'run --rm'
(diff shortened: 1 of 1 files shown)


stsewd (approved):
I wanted to add this, but ended up using the `docker.compose` command. Also, if you don't want to execute everything, you can try `inv docker.compose 'start web'` and `inv docker.compose 'exec web bash'` (I do this when running the tests)

agjohnson:
Honestly, we probably should have wrappers for things like `docker.collectstatic` and running tests so that we don't have 8 competing methods of accomplishing the same task :slightly_smiling_face:

humitos (approved):

humitos:
I like how it is wrapped right now with the high level commands: `.compose` and `.manage`. Those allow you to run custom things without creating another invoke task like `docker.shell_plus` or `docker.dbshell` or whatever.

For running tests, we already have `docker.tests`.

agjohnson:
Yeah, i'd suggest adding `docker.collectstatic` and friends as additional tasks, not replace the high level tasks. It was not clear how i was supposed to run commands like collectstatic, had to ask, and still obviously came up with a different solution than everyone else. Less ambiguity here would make this a bit more approachable, but is something we can tackle later.

PR #77, opened 2020-11-02
Make `--running` the default argument on some commands

### LINKED ISSUE #76: Flip running flag for some/all docker tasks
One thing that confused me about the docker tasks when i was starting was that the tasks usually all spun up a new instance of `web`/whatever to run taks like `manage` and `shell`. I expected the commands to run in an existing container and it took me a long time to hunt down several bugs because these commands were executed in a fresh container.

One issue that the team will hit is that if you are running the instance with `inv docker.up --ext-theme`, you can't run `inv docker.manage collectstatic` to collect assets, as the new instance that is spun up doesn't have the theme enabled and the new assets won't be collected.

It seems like running commands in a running container should be the default, not the exception. Running commands in a fresh container is used to run commands like manage commands without an instance running, which seems edge case to me.

What does the rest of the team do here? I always run an instance if I am doing development. It would be safe for me to always flip the running flag to `True` on all invoke tasks.

humitos:
Now, when calling `inv docker.shell` (and others) all the commands will be
executed in the container that it's already running (using all the context
already specified, like `--webpack` or `--ext-theme`). For this to work, it's
required to have called `inv docker.up` previously, otherwise the command will
fail saying:

ERROR: No container found for web_1

If you still want to run the command anyways, you can add `--no-running`
argument to the invoke command to spin up a new container just to run that
command.

Closes #76

### DIFF
--- a/dockerfiles/tasks.py
+++ b/dockerfiles/tasks.py
@@ -14,12 +14,16 @@ def build(c, cache=False):
     cache_opt = '' if cache else '--no-cache'
     c.run(f'{DOCKER_COMPOSE_COMMAND} build {cache_opt}', pty=True)
 
-@task
+@task(help={
+    'command': 'Command to pass directly to "docker-compose"',
+})
 def compose(c, command):
     """Pass the command to docker-compose directly."""
     c.run(f'{DOCKER_COMPOSE_COMMAND} {command}', pty=True)
 
-@task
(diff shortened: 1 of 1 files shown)


agjohnson (approved):
Looks good! It seems the majority of the team uses the commands like this, so this should be safe

PR #78, opened 2020-11-16
Backup PostgreSQL database when running `migrate`

humitos:
When calling `inv docker.manage migrate` we can pass `--backupdb` to save an SQL
file with the content of the database before running the migration.

The name of the file uses the current date plus the git commit. This allows us
to go back to the previous state in case it's needed.

### DIFF
--- a/dockerfiles/tasks.py
+++ b/dockerfiles/tasks.py
@@ -81,7 +81,7 @@ def shell(c, running=True, container='web'):
 @task(help={
     'command': 'Command to pass directly to "django-admin" inside the container',
     'running': 'Execute "django-admin" in a running container',
-    'backupdb': 'Backup postgres database when running Django "migrate" command',
+    'backupdb': 'Backup postgres database before running Django "manage" command',
 })
 def manage(c, command, running=True, backupdb=False):
     """Run manage.py with a specific command."""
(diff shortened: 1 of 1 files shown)


dockerfiles/tasks.py
```diff
@@ -81,12 +81,17 @@ def shell(c, running=True, container='web'):
...
+    'backupdb': 'Backup postgres database when running Django "migrate" command',
 })
-def manage(c, command, running=True):
+def manage(c, command, running=True, backupdb=False):
     """Run manage.py with a specific command."""
     subcmd = 'run --rm'
     if running:
         subcmd = 'exec'
+
+    if command.startswith('migrate') and backupdb:
```
ericholscher, line 92: Is there a reason to check for `migrate` here? We can backup the DB on another command if we want.

ericholscher (approved):
Curious how well this works, but I've definitely corrupted my DB multiple times this way.

I think a better solution is probably a standard set of fixture data for us all to use in development, and making that easy to load on a fresh DB, but this seems ok for now.

dockerfiles/tasks.py
humitos: Yes. Originally I used `backupdb=True` as default because I wanted to backup every time we run `migrate`. Now that the default is `False`, we can backup on every `manage` command if we want.

We can just remove the `migrate` from here and keep the default as it is now: `False`. Yes.

humitos: I removed the `.startswith('migrate')` in a new commit.

PR #80, opened 2021-01-08
Use S3 as storage backend

humitos:
Required by https://github.com/readthedocs/readthedocs.org/pull/7812


agjohnson:
Any additional steps required here? I assume we'll all need to update common and do a `docker.up --init`?

humitos:
I think I already pin the common commit in the .org/.com repositories, so may not be needed to update.

`docker.up --init` is not required because the buckets are created automatically by django-storages. However, this changed in the latest version of it and we will require to add `create_container.sh` as we had with Azure once we upgrade.

For now, these buckets are automatically created by the app, but we need to give Read permissions to them so they can be accessed from outside. See https://github.com/readthedocs/readthedocs.org/pull/7812/files#diff-6c8480497c5d6d42989c816b96e6ff06b117a03fd891744b565c01a6f92046c8R109

stsewd (approved):
We shouldn't change the commit from the submodule to one that isn't on master, it's confusing, and disappears once this branch is deleted.

PR #81, opened 2021-01-17
Use `CTRL-p CTRL-p` to detach an attached container

humitos:
When attaching a container with:

inv docker.attach web

you can now detach it without killing the process with `CTRL-p CTRL-p`. The
default sequence is CTRL-p CTRL-q for containers running with `-i -t` but it
seems that CTRL-q does not get recognized by Invoke for some reason. So, I'm
using CTRL-p CTRL-p for this.

docs.docker.com/engine/reference/commandline/attach


stsewd (approved):
I wouldn't mind having this default to ctrl-c. Doesn't see to be useful to kill an attached container.

humitos:
Could be. I still prefer something different than CTRL-c tho, since it means something different than detach in my mind.

PR #89, opened 2021-07-12
storage: pin minio version

humitos:
Latest MinIO version requires some changes to use `--console-address`. Also, the
new version of MinIO will require to update the documentation from our
installation guide.

Let's pin it for now to fix the immediate issue and come back with the proper solution.

Solution taken from @astrojuanlu's comment https://github.com/readthedocs/readthedocs.org/issues/8332#issuecomment-878012472


stsewd (approved):
weird versioning convention :D

PR #90, opened 2021-07-12
storage: upgrade MinIO to its latest version

humitos:
Following #89 to upgrade MinIO to its latest version.

Blocked because it requires updating the documentation from https://docs.readthedocs.io/en/latest/development/install.html with the new steps required since the MinIO Console has changed its look&feel.

### DIFF
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -149,7 +149,7 @@ services:
       readthedocs:
 
   storage:
-    image: minio/minio:RELEASE.2021-06-17T00-10-46Z
+    image: minio/minio:RELEASE.2022-02-24T22-12-01Z.fips
     ports:
       - "9000:9000"
       - "9009:9009"
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -152,11 +152,12 @@ services:
     image: minio/minio:RELEASE.2021-06-17T00-10-46Z
     ports:
       - "9000:9000"
+      - "9009:9009"
     environment:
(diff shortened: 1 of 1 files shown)


stsewd (approved):

astrojuanlu (changes requested):
Can we have docs? 🙃 I can write them myself by pushing to this branch before merging. As far as I understand there's no rush to upgrade MinIO?

humitos:
@astrojuanlu we can! I didn't request a review on this one and marked it as blocked because the MinIO UI has changed considerably and I didn't want to jump into it to learn how to do what we are doing in the new UI.

Note the documentation we have to update is on `readthedocs.org` repository at https://docs.readthedocs.io/en/latest/development/install.html

As you already guessed, there is no rush on merging this PR. I just wanted to open it with the new configs because I've already have read about them.

humitos:
One extra note: we may need to read if there is any step required to "migrate the data". I don't know if the data structure has changed in this new release, but is worth taking a look at that as well.

astrojuanlu:
> One extra note: we may need to read if there is any step required to "migrate the data".

How relevant is this for development installations? I mean: sure, it's a pain in the neck having to recreate some projects. I'm just weighing how much effort do we want to put in documenting that.

humitos:
Yeah, I'm fine dropping the storage and re-creating it honestly. I don't think we need to document it for users and make it public in our docs; I was thinking just about _knowing_ if it's required and if there is something like "execute _this_ command once, and it's done" for the core team only. I'm fine having _that_ command in the description of the PR only.

humitos:
I just checked this locally and there is no need to migrate data.

PR #91, opened 2021-08-02
Use an explicit project name on docker build

ericholscher:
I was getting `naming to docker.io/library/dockerfiles_server` when running build without this.
Which caused the rest of the commands to fail.
Not sure what changed here, but this is explicit and fixes it.

### DIFF
--- a/dockerfiles/tasks.py
+++ b/dockerfiles/tasks.py
@@ -13,7 +13,7 @@ DOCKER_COMPOSE_COMMAND = f'docker-compose -f {DOCKER_COMPOSE} -f {DOCKER_COMPOSE
 def build(c, cache=False):
     """Build docker image for servers."""
     cache_opt = '' if cache else '--no-cache'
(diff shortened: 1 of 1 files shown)


humitos (commented):
By default docker build uses the directory name as prefix. However, if the environment variable `COMPOSE_PROJECT_NAME` is defined (https://github.com/readthedocs/readthedocs.org/blob/master/.env#L1) it will use it.

I've already experimented the problem that you are having here because I installed a different version of `docker-compose` that's not the one pinned in our requirements. The newer version behaves differently and breaks because it does not load the environment variable from `.env`

That said, I'm +1 on making this explicit here and have only one place were this prefix if defined. We can also delete the .env file from our repositories.

We should double-check that the `docker-compose` version we have pinned supports `--project-name` and if not update it.

ericholscher:
Yea, I think I'm using the wrong docker version, and fixed this with an ENV var locally. So I'll close this, and we can handle it with another approach more fully.

PR #92, opened 2021-08-17
prospector.yml: ignore ignore-paths

cclauss:


stsewd:
We can just remove these instead of commenting them, but we should also test this setting in all of our repos that use common

cclauss:
Closing in favor of readthedocs/readthedocs.org#8425

PR #96, opened 2021-11-09
Tasks: remove unused ones

### LINKED ISSUE #95: Remove common tasks.py
We don't seem to be using the implementations for version updating or changelog generation from the `tasks.py` in this repository anymore. It seems these files aren't needed anymore and are replaced by the deploy scripts in `readthedocs-ops/deploy` instead.

@humitos can you verify?

humitos:
Closes #95


agjohnson:
I opened #95 to discuss next steps on removing redundant logic, but would also say that I found it rather hard to make any changes to the release process for community/commercial with this logic moved to the release scripts instead. I found myself wanting this logic in `common` and reused or called by the release script, instead of per-repository logic existing only in the release script. I got overwhelmed trying to coordinate the change between 3 repositories at the same time and gave up on the changes.

@humitos do you recall why this logic had to be moved in the first place?

I hit issues with working on this both around trying to use bump2version and to change up how we generate the changelog.

humitos:
> @humitos do you recall why this logic had to be moved in the first place?

We moved off from common because it was not needed here anymore and it was used _only_ from -ops since we are doing all the "preparing repositories" steps from there.

agjohnson:
Yeah, even though it was only used in release, having this logic close to each individual repository did make it easier to work with. I found myself needing a lot of conditional behavior in the release script to start to test improving our release process on each individual repository. It's fine, but I think we could end up adding something similar to this logic if we want to test bump2version and changelog process changes. We can keep this pretty minor though.

agjohnson (approved):

PR #103, opened 2022-02-07
pylint/prospector: disable cyclic-import check

humitos:
I wasn't able to solve this problem in
https://github.com/readthedocs/readthedocs.org/pull/8815 and I don't see there
is anything in that PR's code producing a Cyclic Import. The code works properly
and I haven't seen an exception on it.

I suppose it may be related with a bug in pylint or something similar 🤷🏼

### DIFF
--- a/prospector.yml
+++ b/prospector.yml
@@ -30,6 +30,12 @@ pylint:
     dummy-variables-rgx: '_$|__$|dummy'
     max-line-length: 100
   disable:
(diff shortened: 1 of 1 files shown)


agjohnson:
It might also be worth replacing this check with a dedicated tool. `pycycle` does similar, and can be configured a bit more. I was able to get `pycycle` to pass on readthedocs/readthedocs.org#8815 by tuning the paths used.

PR #108, opened 2022-04-27
Docker: use dedicated db for telemetry

stsewd:
Ref https://github.com/readthedocs/readthedocs.org/pull/9113

### DIFF
--- a/dockerfiles/entrypoints/postgres.sql
+++ b/dockerfiles/entrypoints/postgres.sql
@@ -1,0 +1,3 @@
+-- Initialization script.
+-- https://github.com/docker-library/docs/blob/master/postgres/README.md#initialization-scripts
+CREATE DATABASE telemetry;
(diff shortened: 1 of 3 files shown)


dockerfiles/entrypoints/postgres.sql
```diff
@@ -0,0 +1,3 @@
+-- Initialization script.
+-- https://github.com/docker-library/docs/blob/master/postgres/README.md#initialization-scripts
```
stsewd, line 2: this is only executed the first time the container is created, if you already have one and don't want to re-create it, just create the db manually from the shell.

ericholscher: This comment should be in the code :)

ericholscher (approved):
Looks simple enough.

humitos:
Why this is not required for `docs_db`?

stsewd:
> Why this is not required for `docs_db`?

what thing?

humitos:
The sql script with the `CREATE DATABASE` sentence.

stsewd:
@humitos the one you declare here is created automatically, but you can't declare more than one

https://github.com/readthedocs/common/blob/baa26aa89229b0a59a9507eb5992e559d37775b1/dockerfiles/docker-compose.yml#L144

PR #109, opened 2022-05-05
Celery: keep cancelled builds cancelled

stsewd:
When a task is revoked, it still needs to be picked up by a worker to "delete it",
in the meantime celery will keep those revoked tasks in memory,
if you restart celery, that list is lost, and the tasks will be executed after a worker picks it up.

Ref https://docs.celeryq.dev/en/stable/userguide/workers.html#persistent-revokes

### DIFF
--- a/dockerfiles/entrypoints/build.sh
+++ b/dockerfiles/entrypoints/build.sh
@@ -2,6 +2,11 @@
 
 ../../docker/common.sh
 
+# When a task is revoked, it still needs to be picked up by a worker to "delete it",
+# in the meantime celery will keep those revoked tasks in memory,
+# if you restart celery, that list is lost, and the tasks will be executed after a worker picks it up.
+# We need to save that state in a file so isn't lost.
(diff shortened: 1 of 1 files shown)


dockerfiles/entrypoints/build.sh
```diff
@@ -2,7 +2,9 @@
 
 ../../docker/common.sh
 
-CMD="python3 -m celery -A ${CELERY_APP_NAME}.worker worker -Ofair -c 1 -Q builder,celery,default,build01,build:default,build:large -l DEBUG"
+CELERY_STATE_DIR=/var/run/celery/
+mkdir -p $CELERY_STATE_DIR
+CMD="python3 -m celery -A ${CELERY_APP_NAME}.worker worker -Ofair -c 1 -Q builder,celery,default,build01,build:default,build:large -l DEBUG --statedb=${CELERY_STATE_DIR}worker.state"
```
ericholscher, line 7: We should add a comment here explaining this. Basically what you put in the PR description.

ericholscher (approved):

PR #144, opened 2022-06-09
Use mc instead of aws

benjaoming:
* Use the `mc` MinIO client instead of `aws`, since MinIO policy changes are not supported through aws client.
* Add `wait-for-it.sh` for `web` (waiting for `database`, conditionally waiting for `storage`) and `celery` (conditionally waiting for `search`)
* Remove `wait_for_search.py`

Fixes:

* bucket policies not created, had to be edited manually after creation.


dockerfiles/docker-compose.yml
```diff
@@ -162,3 +161,24 @@ services:
...
+      /bin/sh -c "
+      /usr/bin/mc alias set myminio http://storage:9000 admin password;
+      /usr/bin/mc mb myminio/static;
+      /usr/bin/mc policy set public myminio/static;
+      /usr/bin/mc mb myminio/media;
+      /usr/bin/mc policy set public myminio/media;
+      /usr/bin/mc mb myminio/build-tools;
+      /usr/bin/mc policy set public myminio/build-tools;
+      exit 0;
+      "
```
humitos, line 182: Why not use the same script that we had for this, `createbucket.sh`, to keep consistency with the rest of the operations?

dockerfiles/docker-compose.yml
```diff
@@ -162,3 +161,24 @@ services:
     networks:
       readthedocs:
     command: ["server", "--console-address", ":9009", "/data"]
+    
+  createbuckets:
```
humitos, line 165: When this `createbuckets` will be executed? Every time the `docker-compose up` is run?

If that's the case, we avoided to do that by using the `INIT` variable, so it's run only the first time the environment is setup (passing `--init`)

humitos (commented):
> This was the reason why bucket policies had to be edited manually after creation.

Did you need to edit anything manually? AFAIK, we removed this requirement when we migrated to the usage of `aws` to create the buckets and this is now done automatically when using `--init` the first time.

See https://github.com/readthedocs/readthedocs.org/commit/af217d093f41b4081ff4145ec2a597b122a40ff5

If possible, I'd prefer to not change how this works and solve the issue with the permissions if that's an issue at all. One of the benefits of MinIO is that it's compatible with the AWS API and we can use standard tools with it --if we change the MinIO backend for another that's compatible with AWS API we won't need to upgrade our code, that's :100:

dockerfiles/docker-compose.yml
benjaoming: `createbuckets.sh` was executed from the `web` container which doesn't have the mc client.

dockerfiles/docker-compose.yml
benjaoming: Should add a check on `$INIT` :+1:  Currently, it just runs the commands exits anyways since the buckets already exist.

benjaoming:
I found that the old script was running fine, but policies were set as "Private". This is likely because MinIO doesn't implement this part of the S3 API.

See: "List of Amazon S3 Bucket API's not supported on MinIO", specifically "BucketACL (Use [bucket policies](https://docs.min.io/docs/minio-client-complete-guide#policy) instead)" here: https://docs.min.io/docs/minio-server-limits-per-tenant.html

dockerfiles/docker-compose.yml
benjaoming: Added the check on `$INIT`.

For the sake of simplicity, I like keeping this entrypoint in `.yml` instead of mounting a separate .sh file.

dockerfiles/docker-compose.yml
benjaoming: Marking as resolved, LMK if you want a script back. I think that this will impose configuration of the script, that's why I just preferred to express the whole script in .yml rather than adding environment variables or CLI params.

dockerfiles/docker-compose.yml
benjaoming: The entrypoint script got so bloated that it's now back :)

benjaoming:
I've now added a call to `wait-for-it.sh` because the container all to frequently started executing its entrypoint before the MinIO `storage` container was ready.

dockerfiles/docker-compose.yml
```diff
@@ -162,3 +161,18 @@ services:
     networks:
       readthedocs:
     command: ["server", "--console-address", ":9009", "/data"]
+
+  createbuckets:
+    image: minio/mc
+    volumes:
+      - ${PWD}/common/dockerfiles/entrypoints/createbuckets.sh:/createbuckets.sh
+      - ${PWD}/common/dockerfiles/entrypoints/wait-for-it.sh:/wait-for-it.sh
```
humitos, line 169:
```suggestion
      - ${PWD}/common/dockerfiles/entrypoints/createbuckets.sh:/usr/src/app/docker/createbuckets.sh
      - ${PWD}/common/dockerfiles/entrypoints/wait-for-it.sh:/usr/src/app/docker/wait-for-it.sh
```

To keep consistency with the other containers.

dockerfiles/entrypoints/createbuckets.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# Exit if we are not initializing the setup
+test "${INIT}" || exit 0
+
+/wait-for-it.sh storage:9000 --timeout=30
+/usr/bin/mc alias set myminio http://storage:9000 admin password;
```
humitos, line 7:
```suggestion
/usr/bin/mc alias set myminio http://storage:9000 $MINIO_ROOT_USER $MINIO_ROOT_PASSWORD;
```

dockerfiles/docker-compose.yml
humitos: If possible, I'd like to avoid "spinning up an extra docker container each time to do nothing". We could execute the script from `web` container instead and check for `INIT` as we are already doing.

I think you mentioned somewhere else that `mc` is not available on `web`. However, it seems the installation of `mc` client is pretty simple, just downloading a binary file: https://docs.min.io/minio/baremetal/reference/minio-mc.html#quickstart

humitos: However, probably the simplest way to do this is to re-use the container `storage` and use a bash script as entry point:

```
if INIT:
  # create containers
  mc mb ...
  mc policy ...


server --console-address :9009 /data
```

Looks simpler and cleaner. Also avoids spinning a container that does nothing.

dockerfiles/entrypoints/createbuckets.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# Exit if we are not initializing the setup
+test "${INIT}" || exit 0
+
+/wait-for-it.sh storage:9000 --timeout=30
```
humitos, line 6: I think we already had a "wait for it" command...

Found it! It's called `wait_for_search.py` (https://github.com/rtfd/readthedocs-common/blob/baa26aa89229b0a59a9507eb5992e559d37775b1/dockerfiles/scripts/wait_for_search.py#L1). If we are sold on this new `wait-for-it.sh`, we should replace our custom one with this one to keep consistency.

humitos (changes requested):
Looks good to me!

I'd like to not spin up an extra container for creating buckets if possible. I'm proposing a simpler solution by re-using the `storage` one and following the pattern we already have for other containers (like `web`)

Besides that, I left some other small suggestions as well.

dockerfiles/docker-compose.yml
benjaoming: I'd prefer spinning up a container that's made only to run the `mc` client, rather than spinning up a container that has a different purpose. Don't you think that's better?

It starts and exits in parallel with everything else and nothing depends on it, so it's not really slowing anything down.

humitos: Yes. I think it's better to use the `storage` container to create the buckets the first time it's called.

I understand your point, but I think it's better to keep things simple in this case. We are already doing this on `web`, for example, to run migrations, create super users, collectdata, etc. So, I prefer to follow the same pattern here as well.

dockerfiles/entrypoints/createbuckets.sh
benjaoming: Thanks! I added this with quotes :)

dockerfiles/entrypoints/createbuckets.sh
benjaoming: I can do that, yes

The idea of `wait-for-it.sh` is to have as few requirements as possible AFAIK. It also exists as a `debian` package so not all images will necessarily need to include the shell script.

benjaoming: `wait_for_search.py` is now replaced with `wait-for-it.sh`

`wait_for_search.py` might also have waited indefinitely since there was no `timeout=...` passed to `requests.get()` -- I found that it's necessary to wait for quite a long time for `search` to be ready... ~ 1 minute on my setup.

dockerfiles/docker-compose.yml
humitos: Ohh! You are right! We can't do it from the `storage` container. However, we could do it from `web` in case it's easy to install `mc` there.

Taking a look at the docs, it seems it's pretty simple. Just downloading a binary file: https://docs.min.io/docs/minio-client-quickstart-guide.html (section GNU/Linux)

benjaoming: Trying out the `mc` client in the `web` container now :heart:

benjaoming: It works and has been add to the PR now :tada:

dockerfiles/docker-compose.yml
benjaoming: We opted to use the `web` container to host the `mc` client and this is now added to the PR :+1:

dockerfiles/docker-compose.yml
```diff
@@ -98,7 +101,7 @@ services:
     tty: true
     networks:
       readthedocs:
-    command: ["../../docker/celery.sh"]
+    command: ["../../docker/wait-for-it.sh", "search:9200", "--timeout=360", "--", "../../docker/celery.sh"]
```
humitos, line 104: I think this may be a problem. What happens when we call `inv docker.up --no-search`? I guess it will wait and timeout before running `celery.sh`, right?

The script we used before handles this properly by taking care of `--no-search` checking ` if 'SEARCH' in os.environ` before waiting for search.

humitos (commented):
Looks good to me! We need to handle the "wait for it" for the Celery container when we are using `--no-search` before merging, I guess.

dockerfiles/docker-compose.yml
benjaoming: Ay, thank you for catching that, will fix!

dockerfiles/docker-compose-search.yml
```diff
@@ -8,12 +8,12 @@ volumes:
...
-    environment:
-      - SEARCH=t
+    command: ["../../docker/wait-for-it.sh", "search:9200", "--timeout=360", "--strict", "--", "../../docker/celery.sh"]
 
   web:
-    environment:
-      - SEARCH=t
+    # This appends two blocking wait-for-it.sh calls to one another and if both
+    # are successful, runs the web.sh entrypoint
+    command: ["../../docker/wait-for-it.sh", "search:9200", "--timeout=360", "--strict", "--", "../../docker/wait-for-it.sh", "database:5432", "--timeout=60", "--strict", "--", "../../docker/web.sh"]
```
humitos, line 16: This is getting too complex, IMO. I don't really understand how these chained wait-for-it works and how it interacts with the `--no-search`.

Besides, we are removing the `SEARCH=t` variable from here which is required by the application at https://github.com/readthedocs/readthedocs.org/blob/fe0c723c9bfb7f2330b4f064ebbc2de76a807911/readthedocs/settings/docker_compose.py#L62

In fact, we _don't really require ES to be available **before** starting Celery or Web_. The only case when this is needed is when there is a Celery task from a previous execution. Web container doesn't require ES to startup... it's only required when the user performs a search.

At this point, I think it's better and simple to revert this change and use the `wait_for_search.py` script we used to have. I'm not opposed to this change, but we can introduce wait-for-it in another PR. For now, let's keep this PR scoped for the change of `mc` instead of `aws` which is what we want immediately.

humitos (changes requested):
Changes look good! However, I think this PR deviates too much from the original purpose and got too complex due to the introduction of wait-for-it --which does not seem strictly required.

I'd prefer if we keep this PR simple and only introduce the required changes for `mc` which was the original purpose. I'm sorry 🙏🏼  if I added confusion in my review comments, but originally it looked like a straightforward change and it ended up not being trivial and breaking search on the application.

dockerfiles/docker-compose-search.yml
benjaoming: I can easily change this to not affect the entrypoint command -- it was more of an attempt to bring the behavior out in the docker compose configuration - so have it wait for other containers in the same layer that configures these containers.

benjaoming: ....but looking at the end result, I agree that the chained wait-for-it command wasn't very nice.

I completely missed how SEARCH was used further down in the stack. Thanks for pointing that out :+1:

dockerfiles/docker-compose.yml
benjaoming: This is fixed and tested that `celery` comes up immediately with `--no-search`

benjaoming:
Tested that containers are waiting for each other and `--no-search` option works :heavy_check_mark: 
Rewrote the commit history :heavy_check_mark:

dockerfiles/entrypoints/celery.sh
```diff
@@ -2,7 +2,11 @@
 
 ../../docker/common.sh
 
-python3 ../../docker/scripts/wait_for_search.py
+# If SEARCH is enabled, the celery container should wait for the search
+# container to be ready
+if [ "$SEARCH" != "" ]; then
+  ../../docker/wait-for-it.sh search:9200 --timeout=360 --strict
```
humitos, line 8: 6 minutes? That seems like a _lot_ 🤯

dockerfiles/entrypoints/web.sh
```diff
@@ -4,11 +4,11 @@
 
 if [ -n "$INIT" ];
 then
+    ../../docker/createbuckets.sh
     echo "Performing initial tasks..."
```
humitos, line 8:
```suggestion
    echo "Performing initial tasks..."
    ../../docker/createbuckets.sh
```

humitos (approved):
Looks good to me! 💯 

Pinging @stsewd who is the other person heavily using the Docker development setup in case he has opinions on this.

dockerfiles/entrypoints/celery.sh
benjaoming: Mine just took 121 seconds to start, so I didn't really know what would be a good timeout.. perhaps more like 240 seconds? :)

dockerfiles/docker-compose.yml
```diff
@@ -98,6 +101,7 @@ services:
     tty: true
     networks:
       readthedocs:
+    # NB! Overwritten by docker-compose-search.yml
```
stsewd, line 104: Is this note relevant?

dockerfiles/entrypoints/createbuckets.sh
```diff
@@ -0,0 +1,15 @@
...
+test "${INIT}" || exit 0
+
+/usr/src/app/docker/wait-for-it.sh storage:9000 --timeout=30 --strict
+/usr/bin/mc alias set myminio http://storage:9000 "$MINIO_ROOT_USER" "$MINIO_ROOT_PASSWORD";
+/usr/bin/mc mb myminio/static;
+/usr/bin/mc policy set public myminio/static;
+/usr/bin/mc mb myminio/media;
+/usr/bin/mc policy set public myminio/media;
+/usr/bin/mc mb myminio/build-tools;
+/usr/bin/mc policy set public myminio/build-tools;
```
stsewd, line 13: I think build-tools is a private bucket on production

dockerfiles/entrypoints/createbuckets.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Exit if we are not initializing the setup
+test "${INIT}" || exit 0
+
+/usr/src/app/docker/wait-for-it.sh storage:9000 --timeout=30 --strict
+/usr/bin/mc alias set myminio http://storage:9000 "$MINIO_ROOT_USER" "$MINIO_ROOT_PASSWORD";
+/usr/bin/mc mb myminio/static;
+/usr/bin/mc policy set public myminio/static;
+/usr/bin/mc mb myminio/media;
+/usr/bin/mc policy set public myminio/media;
```
stsewd, line 11: this bucket should be private on .com

stsewd (commented):
Just realized that we were expecting to create all buckets as public before this change, but I think we should try to match production and set some as private. But I'm fine doing that in another PR.

dockerfiles/entrypoints/createbuckets.sh
benjaoming: https://github.com/readthedocs/common/issues/149

dockerfiles/entrypoints/createbuckets.sh
benjaoming: https://github.com/readthedocs/common/issues/149

dockerfiles/entrypoints/celery.sh
benjaoming: Reduced to 180 secs

humitos:
I just realized that we can remove this Python dependency now that it's not gonna be used anymore, https://github.com/readthedocs/readthedocs.org/blob/51de8840cc23a62150c9ad1ac6f1999b87402fd6/requirements/docker.txt#L22-L23

PR #145, opened 2022-06-10
Set COMPOSE_PROJECT_NAME when invoking docker-compose

benjaoming:
Fixes https://github.com/readthedocs/readthedocs.org/issues/9318

### DIFF
--- a/dockerfiles/tasks.py
+++ b/dockerfiles/tasks.py
@@ -1,3 +1,4 @@
+import os
 from invoke import task
 
 DOCKER_COMPOSE = 'common/dockerfiles/docker-compose.yml'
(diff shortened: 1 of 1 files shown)


benjaoming:
Reminder to update https://github.com/readthedocs/readthedocs.org/pull/9319 after this is merged

humitos (changes requested):
I commented on https://github.com/readthedocs/readthedocs.org/pull/9319#discussion_r895558715 about how to solve this without introducing this change.

I prefer if we can read this automatically from the `.env` file so we can keep the current setup working together with corporate Docker instances.

benjaoming:
Re:

> I prefer if we can read this automatically from the .env file so we can keep the current setup working together with corporate Docker instances.

I'll make a check to see if it works, but did you notice that the patch does not propose to overwrite existing env variables?

```python
DOCKER_COMPOSE_PROJECT_NAME = os.environ.get('COMPOSE_PROJECT_NAME', 'community')
```

humitos:
Yes. It seems this PR is not required anymore now that #148 got merged.

benjaoming:
The `.env` file does not trump existing environment variables.

I found a segment in the Compose docs that specifices that `.env` can be used _both_ for container instance variables and any other configuration, so I suppose it means that `COMPOSE_PROJECT_NAME` is supposed to be an intentional part of `.env` (I previously read the docs as if it was not intended for comfiguring Compose itself).

> The “.env” file
>
> You can set default values for any environment variables referenced in the Compose file, or used to configure Compose, in an [environment file](https://docs.docker.com/compose/env-file/) named .env. The .env file path is as follows:

PR #146, opened 2022-06-10
Have same caching defaults as docker-compose build

benjaoming:
If we run just `docker-compose build`, it uses caching. But if we run `inv docker.build` you get no caching. I found that a bit annoying as it got me stuck a couple of times in very long build processes by mistake.

### DIFF
--- a/dockerfiles/tasks.py
+++ b/dockerfiles/tasks.py
@@ -8,11 +8,11 @@ DOCKER_COMPOSE_OVERRIDE = 'docker-compose.override.yml'
 DOCKER_COMPOSE_COMMAND = f'docker-compose -f {DOCKER_COMPOSE} -f {DOCKER_COMPOSE_OVERRIDE} -f {DOCKER_COMPOSE_SEARCH} -f {DOCKER_COMPOSE_WEBPACK}'
 
 @task(help={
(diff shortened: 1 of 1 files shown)


humitos (changes requested):
He, yeah. This may be confusing if you are a heavy user of `docker-compose`. However, This was done like this on purpose, because it's easy to modify a `requirements.txt` file (or others that are not directly changed on the `Dockerfile`) and because the Docker layer is cached, that doesn't get reflected in the resulting image.

So, to avoid knowing too much about the Read the Docs infrastructure, we preferred to not cache by default. However, if you know exactly what you are doing, you can pass `--cache` to force using the cache.

benjaoming:
Ah, I see now why this could be a harmful default even in a development environment, especially when a lot of customized tests have to be run locally. One of the biggest concerns with Docker's cache is about reproducibility - we cannot guarantee to reproduce a build if a non-pinned dependency occurs from the installed set of packages listed in `requirements.txt`. So for example if `my-pinned-package==1.2.3` depends on `some-other-transient-dependency-package>=5.0` then we can end up testing one version on an old cache in development and if production always builds on a clean environment, it can get a different results. It can actually get a very different result, because developers are notoriously lazy so if they start caching their environments, they might have the cache for a very long time :)

PR #147, opened 2022-06-13
Translations: define `inv docker.translations` task

humitos:
This task will be used to automatically update Django locales from
Transifex (`inv docker.translations --action pull`) and
to push updates from the sources to Transifex as well (`inv docker.translations
--action push`)

Related https://github.com/readthedocs/readthedocs.org/pull/9326

### DIFF
--- a/dockerfiles/tasks.py
+++ b/dockerfiles/tasks.py
@@ -176,15 +176,9 @@ def translations(c, action):
         print('You need to export TRANSIFEX_TOKEN environment variable.')
         sys.exit(1)
 
-    if sys.platform in ('linux', 'linux2'):
-        download_file = 'https://github.com/transifex/cli/releases/download/v1.1.0/tx-linux-amd64.tar.gz'
-    elif sys.platform == 'darwin':
-        download_file = 'https://github.com/transifex/cli/releases/download/v1.1.0/tx-darwin-amd64.tar.gz'
-    else:
-        print(f'Platform ({sys.platform}) not supported.')
-        sys.exit(1)
 
     # Install Transifex Client in Docker container
+    download_file = 'https://github.com/transifex/cli/releases/download/v1.1.0/tx-linux-amd64.tar.gz'
     c.run(f'{DOCKER_COMPOSE_COMMAND} run --rm web /bin/bash -c "curl --silent --location {download_file} | tar --extract -z --file=- tx"', pty=True)
 
     if action == 'pull':
--- a/dockerfiles/tasks.py
+++ b/dockerfiles/tasks.py
@@ -179,7 +179,7 @@ def translations(c, action):
 
     # Install Transifex Client in Docker container
     download_file = 'https://github.com/transifex/cli/releases/download/v1.1.0/tx-linux-amd64.tar.gz'
-    c.run(f'{DOCKER_COMPOSE_COMMAND} run --rm web /bin/bash -c "curl --silent --location {download_file} | tar --extract -z --file=- tx"', pty=True)
+    c.run(f'{DOCKER_COMPOSE_COMMAND} run --rm web /bin/bash -c "curl --location {download_file} | tar --extract -z --file=- tx"', pty=True)
 
     if action == 'pull':
         c.run(f'{DOCKER_COMPOSE_COMMAND} run --rm web ./tx --token {transifex_token} pull --force', pty=True)
--- a/dockerfiles/tasks.py
+++ b/dockerfiles/tasks.py
@@ -8,7 +8,7 @@ DOCKER_COMPOSE_SEARCH = 'common/dockerfiles/docker-compose-search.yml'
 DOCKER_COMPOSE_WEBPACK = 'common/dockerfiles/docker-compose-webpack.yml'
 DOCKER_COMPOSE_ASSETS = 'dockerfiles/docker-compose-assets.yml'
 DOCKER_COMPOSE_OVERRIDE = 'docker-compose.override.yml'
-DOCKER_COMPOSE_COMMAND = f'docker-compose -f {DOCKER_COMPOSE} -f {DOCKER_COMPOSE_OVERRIDE} -f {DOCKER_COMPOSE_SEARCH} -f {DOCKER_COMPOSE_WEBPACK}'
+DOCKER_COMPOSE_COMMAND = f'docker-compose -f {DOCKER_COMPOSE_OVERRIDE} -f {DOCKER_COMPOSE} -f {DOCKER_COMPOSE_SEARCH} -f {DOCKER_COMPOSE_WEBPACK}'
 
 @task(help={
     'cache': 'Build Docker image using cache (default: False)',
--- a/dockerfiles/tasks.py
+++ b/dockerfiles/tasks.py
@@ -177,16 +177,17 @@ def translations(c, action):
         sys.exit(1)
 
 
-    # Install Transifex Client in Docker container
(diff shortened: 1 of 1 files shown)


dockerfiles/tasks.py
```diff
@@ -157,3 +160,39 @@ def buildassets(c):
...
+
+    transifex_token = os.environ.get('TRANSIFEX_TOKEN', None)
+    if not transifex_token:
+        print('You need to export TRANSIFEX_TOKEN environment variable.')
+        sys.exit(1)
+
+    if sys.platform in ('linux', 'linux2'):
+        download_file = 'https://github.com/transifex/cli/releases/download/v1.1.0/tx-linux-amd64.tar.gz'
+    elif sys.platform == 'darwin':
+        download_file = 'https://github.com/transifex/cli/releases/download/v1.1.0/tx-darwin-amd64.tar.gz'
```
agjohnson, line 182: Isn't this file just used from inside the Docker container? It seems to me that we don't need to support Mac OS here if so, the platform should always be Linux

dockerfiles/tasks.py
```diff
@@ -1,11 +1,14 @@
...
+
 from invoke import task
 
 DOCKER_COMPOSE = 'common/dockerfiles/docker-compose.yml'
 DOCKER_COMPOSE_SEARCH = 'common/dockerfiles/docker-compose-search.yml'
 DOCKER_COMPOSE_WEBPACK = 'common/dockerfiles/docker-compose-webpack.yml'
 DOCKER_COMPOSE_ASSETS = 'dockerfiles/docker-compose-assets.yml'
 DOCKER_COMPOSE_OVERRIDE = 'docker-compose.override.yml'
-DOCKER_COMPOSE_COMMAND = f'docker-compose -f {DOCKER_COMPOSE} -f {DOCKER_COMPOSE_OVERRIDE} -f {DOCKER_COMPOSE_SEARCH} -f {DOCKER_COMPOSE_WEBPACK}'
+DOCKER_COMPOSE_COMMAND = f'docker-compose -f {DOCKER_COMPOSE_OVERRIDE} -f {DOCKER_COMPOSE} -f {DOCKER_COMPOSE_SEARCH} -f {DOCKER_COMPOSE_WEBPACK}'
```
stsewd, line 11: this change isn't related to this PR

humitos: I've done it because I need it to when calling the deploy script. Otherwise, the compose project is not read properly from .env file.

dockerfiles/tasks.py
humitos: Oh, yeah. You are right. I was trying a different approach first and then moved the code to here. I will remove this now.

agjohnson:
I'm having trouble picturing how this all will fit together with the release process.

It seems like executing inside Docker during release is going to make this more complicated. In the deploy task, we are already checking out the repositories to a temporary path -- or executing the deploy task inside Docker itself.

So to run the translation commands from the temporary path, it seems a new instance would spin up from the temporary path. And what does execution of these tasks from inside an already running Docker container for the `deploy` task look like?

dockerfiles/tasks.py
stsewd: As mentioned in https://github.com/readthedocs/readthedocs.org/pull/9319#discussion_r895914856, this change will break other things

humitos:
There are other PRs involved about this work (all of them prefixed with "Translations").

We need to run these commands from inside Docker because we need the Python environment with all the packages installed (note we are running `django-admin` commands). In https://github.com/readthedocs/readthedocs-ops/pull/1202 you will see how we call these tasks from the `deploy.sh` command. It's pretty simple, we just `cd` into the temporary path we just cloned and run `invoke` from there. All the resulting modified files (.po/.mo) are saved into that temporary path, committed, and push properly. Similar to what we are doing with the changelog file.

dockerfiles/tasks.py
humitos: Cool! We can undo it before merging this PR, but we need to fix the `.env` issue for this PR to work.

agjohnson:
I understood the technical execution of the code, but my questions were more about adding Docker execution to our release process:

> So to run the translation commands from the temporary path, it seems a new instance would spin up from the temporary path.

It does not. Only the cache, db, and storage containers are created. All removed afterwards due to `--rm` flag.

> And what does execution of these tasks from inside an already running Docker container for the deploy task look like?

I don't execute our `ops/deploy` tasks using Docker, though I know that was the original intention. Testing this quickly, it works okay for this execution pattern.

But if anyone is still executing the deploy task using Docker, these tasks would be executing Docker compose commands from inside a Docker container.

dockerfiles/tasks.py
```diff
@@ -157,3 +160,33 @@ def buildassets(c):
...
+def translations(c, action):
+
+    if action not in ('pull', 'push'):
+        print(f'Action passed ("{action}") not supported. Use "pull" or "push".')
+        sys.exit(1)
+
+    transifex_token = os.environ.get('TRANSIFEX_TOKEN', None)
+    if not transifex_token:
+        print('You need to export TRANSIFEX_TOKEN environment variable.')
+        sys.exit(1)
```
agjohnson, line 177: This seems like it should be a warning or requirement at the top of the deploy prepare process. Otherwise, the missing variable isn't known until it's too late.

dockerfiles/tasks.py
```diff
@@ -157,3 +160,33 @@ def buildassets(c):
...
+
+
+    # Install Transifex Client in Docker container
+    download_file = 'https://github.com/transifex/cli/releases/download/v1.1.0/tx-linux-amd64.tar.gz'
+    c.run(f'{DOCKER_COMPOSE_COMMAND} run --rm web /bin/bash -c "curl --silent --location {download_file} | tar --extract -z --file=- tx"', pty=True)
```
agjohnson, line 182: Not sure what happened, but this command froze and killed the task when I tried first. With `--silent`, I didn't know what was happening. Is there are reason we want `--silent` here? If curl output doesn't play well with pty output, at least some note that a download was started would be helpful.

agjohnson (approved):
I tried this locally, using `ops/deploy` prepare task on a dry run. It mostly worked, with some hiccups, but I did have to export a number of additional pieces. I'll raise this in the related ops PR.

dockerfiles/tasks.py
humitos: Yeah. I put it here because as this step is optional and can be skipped, I didn't want to fail the release if there is no Transifex token specified. I think there is a tradeoff here. Do you prefer to fail the process at the beginning anyway?

Another option could be using a warning as you mentioned at the start and always skip this step if the token is not provided. However, I'm sure this way it will probably always be skipped :)

humitos: BTW, I think this is fine here, because this command _requires_ this variable. However, we can add extra code to the deploy/release script to make it nicer --but we shouldn't remove this from here.

dockerfiles/tasks.py
humitos: Note to myself: Revert this change once https://github.com/readthedocs/common/pull/148 got merged.

dockerfiles/tasks.py
agjohnson: > However, we can add extra code to the deploy/release script to make it nicer --but we shouldn't remove this from here

Oh yeah, I was thinking this option as well. This argument makes sense here, but probably eventually also some mechanism in the prepare task would be helpful to keep translations on track. Keeping this optional for now seems fine, and perhaps adding a requirement to the prepare task is what makes this step mandatory.

PR #150, opened 2022-06-15
Local development: use `nodemon` instead of `watchmedo`

humitos:
See https://github.com/readthedocs/readthedocs.org/pull/9338


agjohnson (approved):
Easy enough. Do we want to jump right into this, or test this locally for a bit? Anyone having trouble with inotify or watchdog already would be a good test candidate

agjohnson:
You and @stsewd use the local environment the most frequently, I'll defer to you both on that. I'm +1 on just going forward, but discount my vote a bit as I don't work as frequently with my local env.

stsewd (approved):
In worked on my local testing

PR #151, opened 2022-06-17
Development: allow to use --ngrok when starting Docker

humitos:
Together with https://github.com/readthedocs/readthedocs.org/pull/9353 this
commit will allow developers to run NGROK with the local instance just by
passing another argument.

```
inv docker.up --ngrok readthedocs.ngrok.io
```


dockerfiles/docker-compose.yml
```diff
@@ -23,7 +23,7 @@ services:
       - proxito
       - storage
     volumes:
-      - ${PWD}/dockerfiles/nginx:/etc/nginx/conf.d
+      - ${PWD}/dockerfiles/nginx:/etc/nginx/templates
```
benjaoming, line 26: Really good idea to start using envsubst :+1: This is a good step towards convergence between development docker images and production. Ideally, the same Docker images can be used in multiple deployments and after having isolated variables as environment variables, we can reduce reliance on volumes and build configurations straight into the images.

dockerfiles/tasks.py
```diff
@@ -42,8 +42,9 @@ def down(c, volumes=False):
     'webpack': 'Start webpack development server (default: False)',
     'ext-theme': 'Enable new theme from ext-theme (default: False)',
     'scale-build': 'Add additional build instances (default: 1)',
+    'ngrok': 'Start environment to be accessed via NGROK (default: False)',
```
benjaoming, line 45:
```suggestion
    'ngrok': 'Start environment to be accessed via ngrok (default: "") - create an account for ngrok, install the service and use your full ngrok domain, for instance --ngrok=readthedocs.ngrok.io',
```

benjaoming (approved):
LGTM :+1: I tried using it locally, albeit without installing ngrok. But I really like it, it's a great idea!

dockerfiles/docker-compose.yml
benjaoming: Note that this is incompatible with current `readthedocs-corporate` setup.

dockerfiles/tasks.py
humitos: I think we can add this extra paragraph to the documentation on the install development page if I haven't already, to avoid being too verbose in the quick help of the command. However, the example suggestion looks good 👍🏼

humitos:
```suggestion
    'ngrok': 'ngrok domain to serve the application. Example: "17b5-139-47-118-243.ngrok.io"',
```

PR #155, opened 2022-08-17
Gulp eslint: allow es6

stsewd:
This is so we can use arrow functions, let statements, const, and all other goodies that make JS a little better.


agjohnson (approved):
Seems okay. I also don't use eslint anymore, Jest is quite nice. I forget where we are using eslint at all, so might be worth evaluating tools.

The standard toolchain our repos are moving towards is Prettier for linting, Jest for testing, and Webpack + Babel for bundling.

PR #156, opened 2022-10-18
Add `blacken-docs` in pre-commit hook

dojutsu-user:
Closes: https://github.com/readthedocs/readthedocs.org/issues/9502
Related: https://github.com/readthedocs/readthedocs.org/pull/9672

### DIFF
--- a/pre-commit-config.yaml
+++ b/pre-commit-config.yaml
@@ -75,3 +75,9 @@ repos:
         # darker is not compatible with v22.1.0
         # https://github.com/akaihola/darker/issues/264
         - black==21.12b0
+
+- repo: https://github.com/asottile/blacken-docs
+  rev: v1.12.1
+  hooks:
+    - id: blacken-docs
+      additional_dependencies: [black==22.10.0]


benjaoming (approved):
Nicely done and nicely demonstrated! Thanks for this contribution!

Looking over the configuration options for the blacken-docs hook, I think you did a very great choice in making the black version explicit.

So another "LGTM" from me :+1: 

I'll leave it for another team-member to review.

dojutsu-user:
Thank you @benjaoming 😄

pre-commit-config.yaml
```diff
@@ -75,3 +75,9 @@ repos:
         # darker is not compatible with v22.1.0
         # https://github.com/akaihola/darker/issues/264
         - black==21.12b0
+
+- repo: https://github.com/asottile/blacken-docs
+  rev: v1.12.1
+  hooks:
+    - id: blacken-docs
+      additional_dependencies: [black==22.10.0]
```
ericholscher, line 83: Seems this is depending on a different version of black than defined above? We should add a comment explaining why that is, if we're doing it for some reason.

ericholscher (commented):
This looks good, other than my question around versions.

pre-commit-config.yaml
benjaoming: I think each hook is allocated its own virtualenv?

There is a comment for the `darker` hook about `black==21.12b0` -- it has issues with `darker`, but that comment wouldn't apply for `blacken-docs`, so I guess they can be allowed to drift as long as the `darker` hook has this issue? Once the issue is solved, they can align?

PR #166, opened 2023-03-05
pre-commit: run `prospector` from inside `pre-commit`

humitos:
- merge both pospector YAML file
- run prospector from inside pre-commit

Related: https://github.com/readthedocs/readthedocs.org/pull/10105


prospector.yml
```diff
@@ -7,16 +7,39 @@ uses:
...
+  - payments
+  - profiles
+  - projects
+  - redirects
+  - api
+  - rtd_tests
+  - search
+  - settings
+  - templates
+  - vcs_support
```
stsewd, line 38: This will now ignore all warnings for all these modules, we should fix all the errors before merging the prospector files.

humitos: Yes. I copied these ignores from our current configuration. We are already ignoring all these files.

Fixing the errors from all these files is outside the scope of this PR.

humitos: OK. I remove them all. We will be fixing them as we go when touching/modifying those files.

stsewd: So, are going to see new warnings every time we touch existing code only? This is, not in the existing/unchanged code?

humitos: Yes. Unfortunately, prospector does not work over "chunks of codes" as darker. So, it will run over the whole file.

stsewd: I'm fine fixing the errors while changing something else in those files, @ericholscher may have some thoughts here.

humitos: We can blackify the whole code, but I don't think we can lint it. Unfortunately, some linting issues are not solved automatically by any tools 🤷🏼

ericholscher: Ah, true. Good point.  👍

ericholscher (approved):
I think I'm 👍 on this change, but I do worry about going low -> high in strictness. Do we know how annoying this will be? Might be wise to start at medium, but if all our code passes with this config, I think it's fine to ship.

humitos:
I update the strictness to be `medium`, which seems like a good initial compromise 👍🏼  I'm sure this configuration won't be perfect, but we can easily tune it as we go.

I main goal here is to automatically run prospector locally on each commit, so we don't  push code with lint issues to GitHub and we realize after 20 minutes that there was an import that was not used or similar.

ericholscher:
> I main goal here is to automatically run prospector locally on each commit, so we don't push code with lint issues to GitHub and we realize after 20 minutes that there was an import that was not used or similar.

That is a great goal, and I'm 💯 on having linting in pre-commit locally 👍

PR #167, opened 2023-03-06
Development: argument to set logging level

humitos:
It was a little annoying that we had the logging level set to DEBUG since we are flooded wth a lot of messages, even when debugging things.

I set the default logging level to INFO, but added a `log-level=` argument to easily change it in case we want to be more verbose while working locally.

### DIFF
--- a/dockerfiles/tasks.py
+++ b/dockerfiles/tasks.py
@@ -43,13 +43,15 @@ def down(c, volumes=False):
     'ext-theme': 'Enable new theme from ext-theme (default: False)',
     'scale-build': 'Add additional build instances (default: 1)',
     'ngrok': 'ngrok domain to serve the application. Example: "17b5-139-47-118-243.ngrok.io"',
+    'log-level': 'Logging level for the Django application (default: INFO)',
 })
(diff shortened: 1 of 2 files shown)


dockerfiles/tasks.py
```diff
@@ -43,13 +43,15 @@ def down(c, volumes=False):
     'ext-theme': 'Enable new theme from ext-theme (default: False)',
     'scale-build': 'Add additional build instances (default: 1)',
     'ngrok': 'ngrok domain to serve the application. Example: "17b5-139-47-118-243.ngrok.io"',
+    'log-level': 'Logging level for the Django application (default: INFO)',
```
benjaoming, line 46: Possible values are DEBUG, ERROR and WARNING, right?

benjaoming (approved):
Thank you!!! I've been missing this feature, too :heart:

dockerfiles/tasks.py
humitos: Yes. These are the supported values: https://docs.djangoproject.com/en/4.1/topics/logging/#loggers

PR #168, opened 2023-03-08
pre-commit: tweaks after linting our code

humitos:
Related https://github.com/readthedocs/readthedocs.org/pull/10129

### DIFF
--- a/pre-commit-config.yaml
+++ b/pre-commit-config.yaml
@@ -18,12 +18,22 @@ repos:
     - id: debug-statements
     - id: fix-byte-order-marker
     - id: end-of-file-fixer
-      exclude: 'static'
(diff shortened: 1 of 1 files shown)


benjaoming (approved):
No worries about excluding these :+1: 

Does this mean that `pre-commit run --all-files` works?

humitos:
> Does this mean that `pre-commit run --all-files` works?

Not yet. There are plenty of files that don't pass `prospector`. That will take some extra effort to fix because they are not just quick and simple fixes, but we are getting closer.

benjaoming:
Ah yeah, maybe those complicated things can sit around until we decide what to do with Ruff.

humitos:
I never used it, but I've heard about it and looks promising 😄

PR #174, opened 2023-04-18
Remove max-line-length from pep8 because it does not get silenced by ignore-patterns

benjaoming:
Referencing "checks" failure here: https://github.com/readthedocs/readthedocs-corporate/pull/1579

Also on a related note: We should minimize prospector or remove it.


benjaoming:
I can't actually make `ignore-paths` nor `ignore-patterns` ignore any errors that I trigger with Prospector's pylint or pep8.

To me, it just looks like a prospector bug. And it seems to happen on my local env and in Circle CI.

PR #177, opened 2023-05-22
nodemon: monitor readthedocs.org volume

benjaoming:
tested and works :+1: 

Updated: When working with changes to `readthedocs.org` that affect readthedocs-corporate, it's nice to have the mounted volume with `readthedocs.org` also cause reloads. Otherwise, it's only changes to the `readthedocs-corporate` repo that causes reloads.

### DIFF
--- a/dockerfiles/nodemon.json
+++ b/dockerfiles/nodemon.json
@@ -2,7 +2,7 @@
     "verbose": false,
     "delay": 2000,
     "ext": "py",
-    "watch": ["readthedocs", "readthedocsinc"],
+    "watch": ["readthedocs", "readthedocsinc", "../readthedocs.org/readthedocs"],
     "ignore": [".tox/*", ".direnv/*", "user_builds/*", "*/management/commands*", "*migrations/*", "*test*", "*.pyc", "*.pyo", "logs/*"],
     "signal": "SIGTERM"
 }


humitos:
I'm not sure to understand this PR. What's the issue it solves?

humitos:
Good point! 💯

dockerfiles/nodemon.json
```diff
@@ -2,7 +2,7 @@
     "verbose": false,
     "delay": 2000,
     "ext": "py",
-    "watch": ["readthedocs", "readthedocsinc"],
+    "watch": ["readthedocs", "readthedocsinc", "../readthedocs.org/readthedocs"],
```
humitos, line 5: Does this work properly on .org? Doesn't it fails saying this directory doesn't exists there? Should we remove the first `"readthedocs"` to avoid watching the same files twice?

humitos (approved):
Looks good! There are some questions relating how this work on .org, tho.

dockerfiles/nodemon.json
benjaoming: It's in the `common` repo, so it'd have to work on both repos at the same time.

But readthedocs.org never failed with the missing `readthedocsinc` path spec, so it's safe :+1:

benjaoming: Works on .org :+1:

PR #183, opened 2023-07-24
Development: do not install docker-compose

humitos:
Docker Copmose V1 is deprecated and should not be used anymore. The new version, V2, is written in Go and it can't be installed with `pip`.

### DIFF
--- a/dockerfiles/requirements.txt
+++ b/dockerfiles/requirements.txt
@@ -1,3 +1,2 @@
 # requirements file to launch Read the Docs using docker-compose
 invoke==1.7.1
-docker-compose==1.29.2


stsewd:
> The new version, V2, is written in Go and it can't be installed with pip.

The new version is shipped with docker itself (if you install docker normally), we should update our commands to use `docker compose` instead of `docker-compose`

humitos:
I understand that `docker compose` is just an alias to `docker-compose`. If you uninstall `docker-compose` package on Arch `docker compose` doesn't work.

ericholscher (approved):
Seems reasonable. Is there a way for us to specify a version anymore?

humitos:
Hrm, I don't think so, but I guess it will be fine to use the suggested one by the official Docker docs. I suppose we will need to clearly specify this in the docs once V3 comes out and be incompatible probably.

PR #193, opened 2024-01-03
Elasticsearch: disable HTTPS

stsewd:
Instead of disabling the whole security package (https://github.com/readthedocs/common/commit/151504fbc8a2b399a6277c92bf549f3168c026b1) we just need to disable HTTPS for Elasticsearch.

I just hit this problem in a fresh environment.


humitos (approved):
Great!

PR #198, opened 2024-02-10
Remove pydocstyle check that Black doesn't use

ericholscher:
I'm assuming we shouldn't have checks that conflict with Black,
so removing this.


humitos:
pydocstyle adds a blank line and the Black removes it? 🤣

humitos (approved):

ericholscher:
Yea, at least I couldn't figure out how to make it match. Our other code seems to follow the pydocstyle format, so it's possible I might have been doing something wrong?

PR #200, opened 2024-02-14
wrangler: move the Dockerfile to `common/`

humitos:
This file was in community before, but it has to be shared with Read the Docs for Business too.

### DIFF
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -37,7 +37,7 @@ services:
   wrangler:
     build:
       context: ${PWD}
-      dockerfile: ${PWD}/dockerfiles/Dockerfile.wrangler
+      dockerfile: ${PWD}/common/dockerfiles/Dockerfile.wrangler
     volumes:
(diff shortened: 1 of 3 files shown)


dockerfiles/docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
   wrangler:
     build:
       context: ${PWD}
-      dockerfile: ${PWD}/dockerfiles/Dockerfile.wrangler
+      dockerfile: ${PWD}/common/dockerfiles/Dockerfile.wrangler
     volumes:
       - ${PWD}/dockerfiles/force-readthedocs-addons.js:/usr/src/app/docker/force-readthedocs-addons.js
```
stsewd, line 42: Don't we still need this file under .com?

humitos: 🤔 Hrm, good point. I think this file will be shared in `common/` then since it will be the same.

stsewd (approved):

PR #203, opened 2024-02-27
CF worker: remove specific warning admonitions

humitos:
In https://github.com/readthedocs/readthedocs-ops/pull/1429 we made the selector to remove the "external version warning" a lot more specific.

It fixed the original issue, but it resulted in not matching in other themes. This PR adds extra specific CSS selector for Furo and Book themes.

Closes https://github.com/readthedocs/readthedocs.org/issues/11136


agjohnson:
This looked fine, we talked about this a little bit during sprint planning/overview. It got knocked off my notification list apparently.

PR #205, opened 2024-03-05
CF: inject `readthedocs-resolver-filename` HTML meta tag

humitos:
We lost this work when moving the `.js` file from the ops repository into community and then again to common.

This is required to be deployed together with addons 0.10.0.

Related:

* https://github.com/readthedocs/readthedocs.org/pull/11126
* https://github.com/readthedocs/addons/issues/211


humitos:
I've already deployed this since it's required for today's addons deploy 👍🏼

stsewd (approved):

PR #207, opened 2024-03-06
Development: reduce gunicorn loglevel

humitos:
Remove a bunch of these following lines when modifying a file.

```
proxito_1      | [2024-03-06 15:32:58 +0000] [22] [ERROR] Worker (pid:23) was sent SIGTERM!
proxito_1      | [2024-03-06 15:32:58 +0000] [22] [ERROR] Worker (pid:24) was sent SIGTERM!
```

### DIFF
--- a/dockerfiles/entrypoints/proxito.sh
+++ b/dockerfiles/entrypoints/proxito.sh
@@ -2,7 +2,7 @@
 
 ../../docker/common.sh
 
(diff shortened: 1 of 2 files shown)


stsewd (commented):
Feels like we may be missing some important information by setting the log level that high. If this is something getting on the way of debugging a specific problem, I think it's fine to just do it locally.

PR #210, opened 2024-03-08
Search: protect ourselves from executing long queries

humitos:
This is similar to what we are doing in PostgreSQL to avoid running long queries and degrading the service.

Related https://github.com/readthedocs/addons/issues/84

### DIFF
--- a/dockerfiles/docker-compose-search.yml
+++ b/dockerfiles/docker-compose-search.yml
@@ -24,6 +24,7 @@ services:
       - node.name=search
       - cluster.routing.allocation.disk.threshold_enabled=false
       - cluster.info.update.interval=30m
+      - search.default_search_timeout=15s
       - "ES_JAVA_OPTS=-Xms128m -Xmx128m"
       - ELASTIC_PASSWORD=password
       # Disable HTTPS on development.


humitos:
This PR is only for development, but we should open a similar one for production. I haven't found where we configure these settings in production yet, tho.

humitos:
I checked the configs in Elastic Cloud, but I wasn't able to find _where_ to configure this there.

humitos:
@stsewd do you know how to enable this setting in production?

stsewd:
Looks like you can do it from here https://www.elastic.co/guide/en/cloud/current/ec-add-user-settings.html, but that document says that only some options are available. I don't see that option listed there.

ericholscher:
I think we can probably close this?

humitos:
> I think we can probably close this?

This is still useful, but not a priority, tho.


> Looks like you can do it from here [elastic.co/guide/en/cloud/current/ec-add-user-settings.html](https://www.elastic.co/guide/en/cloud/current/ec-add-user-settings.html), but that document says that only some options are available. I don't see that option listed there.

I think that's because those are "User settings".

Here, https://www.elastic.co/guide/en/elasticsearch/reference/current/search-your-data.html#search-timeout, it talks about setting the `search.default_search_timeout` using the [cluster settings API](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-update-settings.html); but I'm not sure how we are calling that API in production 🤷🏼

ericholscher:
If we think it's useful I'm fine to merge it, but I don't think I've hit issues with this at all in dev?

PR #211, opened 2024-03-08
Exclude migrations, vendor, .min.js, .svg, and static/*.js from pre-commit

yarikoptic:
Those files are copied from other sources or edited by other tools and IMHO not subject for pre-commit handling.

Refs:
- https://github.com/readthedocs/readthedocs.org/issues/11200
- https://github.com/readthedocs/readthedocs.org/pull/11199


ericholscher (approved):
This makes sense to me :+1:

PR #213, opened 2024-04-11
Dev: Pass http-domain to RTD_PUBLIC_DOMAIN too

hoyes:
acc68e51 added the ability to customize RTD_PRODUCTION_DOMAIN using the --http-domain argument to docker.up

The Docker Compose settings also allows RTD_PUBLIC_DOMAIN to be configured using an env var but this appears to be unused. Pass http-domain to RTD_PUBLIC_DOMAIN too so that the generated docs can be tested on a custom domain too.

Pass RTD_PUBLIC_DOMAIN to the containers wherever RTD_PRODUCTION_DOMAIN is used.

### DIFF
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -68,6 +68,7 @@ services:
     environment:
       - DOCKER_NO_RELOAD
       - RTD_PRODUCTION_DOMAIN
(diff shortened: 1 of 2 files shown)


stsewd (approved):
Makes sense to me!

PR #214, opened 2024-05-16
CF worker: remove `script[id=READTHEDOCS_DATA] + script`

humitos:
We used to removed this previously, but the CSS selector was broken, so we stopped doing it. Today, I found the correct selector to remove it.

These are the two `script` this commit removes:
https://github.com/readthedocs/readthedocs-sphinx-ext/blob/main/readthedocs_ext/_templates/readthedocs-insert.html.tmpl#L23-L31


humitos:
Hrm... it seems that Cloudflare does not support this selector :/

```
 Parser error: Unsupported combinator `+` in selector. 
```

ericholscher:
Yea, CSS selectors are dark magic :/

PR #215, opened 2024-06-04
Update docker compose files

stsewd:
- Version is no longer needed (https://github.com/docker/compose/issues/11628)
- The version of postgres doesn't work with the latest version of docker for some reason, puling that image results in `layers from manifest don't match image configuration`. Also, 12.17 is what we use in production.

### DIFF
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -1,5 +1,3 @@
-version: '3'
-
 volumes:
   build-user-builds:
   storage_data:
@@ -217,7 +215,7 @@ services:
       readthedocs:
 
   database:
-    image: postgres:12.5
+    image: postgres:12.17
     environment:
       - POSTGRES_USER=docs_user
       - POSTGRES_PASSWORD=docs_pwd
--- a/dockerfiles/docker-compose-search.yml
+++ b/dockerfiles/docker-compose-search.yml
@@ -1,6 +1,5 @@
 # docker-compose-search.yml starts "search" service and override some
 # services to link them so they can use the "search" as well.
-version: '3'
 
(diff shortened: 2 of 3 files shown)


dockerfiles/docker-compose.yml
```diff
@@ -1,5 +1,3 @@
-version: '3'
```
humitos, line 1: I'm getting the same warning message about the version, but I'm not sure we can remove it since we still tell people to use the Docker Compose version that's installed via `pip` in step 3 from our guide: https://dev.readthedocs.io/en/latest/install.html#set-up-your-environment

stsewd: Don't know. All of our scripts no longer use that docker-compose command anyway, we use `docker compose`. We should just remove that dependency.

humitos: We should test if our current guide continues working before making this change, then. If it doesn't, we should update it to tell people to install the latest version of Docker that contains Docker Compose.

stsewd: Don't think we should support the old version of docker. We can just remove that dependency, we already link to the official guide installation of docker, and that works.

humitos: As long as the guide continues working, 👍🏼

stsewd: We are not even installing docker-compose anymore with pip https://github.com/readthedocs/common/commit/9932d8a070a387211bf283dd2ff20d2f42af62ed.

humitos: Cool

humitos (approved):

PR #216, opened 2024-06-13
Use social account env vars on celery and celery-beat instances

agjohnson:
Without this, celery and celery beat fail to refresh oauth tokens for
social account providers. I found this debugging why refresh tokens
failed while I was testing GitLab status updates.


stsewd (approved):

humitos:
It would be good to add a comment above these settings here explaining why they are needed

PR #218, opened 2024-07-09
Cloudflare: skip running the worker if >1Mb

humitos:
Don't execute our worker if the file size is bigger than 1Mb. This is to avoid a "Parser error: The memory limit has been exceeded." for now, but we should be able to find a better solution that also works for these big files.

Closes https://github.com/readthedocs/readthedocs-ops/issues/1513

### DIFF
--- a/dockerfiles/force-readthedocs-addons.js
+++ b/dockerfiles/force-readthedocs-addons.js
@@ -64,6 +64,7 @@ async function handleRequest(request) {
 
   // get the content type of the response to manipulate the content only if it's HTML
   const contentType = originalResponse.headers.get("content-type") || "";
+  const contentLength = originalResponse.headers.get("content-length") || "";
   const injectHostingIntegrations =
     originalResponse.headers.get("x-rtd-hosting-integrations") || "false";
   const forceAddons =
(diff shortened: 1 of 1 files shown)


dockerfiles/force-readthedocs-addons.js
```diff
@@ -73,6 +74,17 @@ async function handleRequest(request) {
...
   console.log(`X-RTD-Force-Addons: ${forceAddons}`);
   console.log(`X-RTD-Hosting-Integrations: ${injectHostingIntegrations}`);
+  console.log(`ContentLength: ${contentLength}`);
+
+  // skip executing the Cloudflare Worker if the response's body is bigger than 1Mb
+  //
+  // the user will see the page rendering properly, but without any of the addons injected
+  // this is an intermediate workaround/solution for
+  // https://github.com/readthedocs/readthedocs-ops/issues/1513
+  if (contentLength && contentLength >= 1000000) {
```
stsewd, line 84: contentLenght is probably a string here, if this is working is because of JS implicit type conversion. But we should cast it as integer.

ericholscher (approved):
Seems worth a try -- wish we could just run the worker without pulling the whole file into memory, but I guess if we're doing string compares on it, we need it in memory. I wonder if we can just do a `head` or similar on it?

humitos:
We could probably do a HEAD request instead of a GET, but that means will be doing two request per file always just to protect us for this edge case. However, we will still need to do the real GET request to be able to return it back to the client. I think we won't be earning anything by doing a HEAD first, actually.

Honestly, it doesn't make sense to me that parsing a 4Mb HTML consumes more than 128Mb, but 🤷🏼 -- in particular, if we are working _only_ over the `head` tag. Maybe we can tune a little more the CSS selector to be based on `head` tag like `head > script[...]` instead of just `script[...]` but I'm not sure if that will help.

humitos:
I just deploy this as a test but it didn't work because it seems the NGINX from production doesn't return `Content-Length`. The version we use in our local development environment, does 😞

humitos:
Yeap, immediately I enabled `gzip on` locally, it stops sending the `Content-Length`...

humitos:
Meh, even calling just `HTMLReWrite().transform(originalResponse)` produces an exception for memory limit. I don't think there is too much we can do here since we are not applying _any transformation_ to the original response. I'm thinking this is probably a bug on CF at this point 🤷🏼 


```
wrangler_1     | ✘ [ERROR] workerd/server/server.c++:2838: error: Uncaught exception: workerd/api/html-rewriter.c++:95: failed: remote.jsg.TypeError: Parser error: The memory limit has been exceeded.
```

humitos:
With the latest version of wrangler, `3.63.2` I was able to reproduce the error in the same way than in production: blank page. With the version we are currently using, it returns the page and sometimes it also adds the addons. So, the old version works better than the new one 🙃

humitos:
I filled a ticket on Cloudflare to ask them if this is the expected behavior.

humitos:
Closing for now as we don't have a good way to solve this issue yet. We will continue the conversation in the issue.

PR #222, opened 2024-09-18
Worker: add HTTP status as meta tag

humitos:
We need to know what's the HTTP status for the original request to be able to decide what addons will be enabled. There are some addons we want to disable on 404s for example.

Related: https://github.com/readthedocs/addons/issues/371


agjohnson (approved):
This approach looks good :+1:

PR #224, opened 2024-10-15
Worker: attempt to catch errors

humitos:
Note that this doesn't work as expected, but I'm not sure why. For some reason what's inside the `catch` block is not executed. It fails with the following error:

```
wrangler_1     | ✘ [ERROR] Uncaught Error: This is an issue
wrangler_1     | 
wrangler_1     |       at element (file:///usr/src/app/docker/force-readthedocs-addons.js:205:11)
wrangler_1     | 

wrangler_1     | ✘ [ERROR] Uncaught (in response) Error: This is an issue
wrangler_1     | 

wrangler_1     | ✘ [ERROR] Uncaught (async) Error: This is an issue
wrangler_1     |
```

Closes #1513

### DIFF
--- a/dockerfiles/force-readthedocs-addons.js
+++ b/dockerfiles/force-readthedocs-addons.js
@@ -86,39 +86,49 @@ async function handleRequest(request) {
   // - content type has to be "text/html"
   // when all these conditions are met, we remove all the old JS/CSS files and inject the new beta flyout JS
 
-  // check if the Content-Type is HTML, otherwise do nothing
-  if (contentType.includes("text/html")) {
-    // Remove old implementation of our flyout and inject the new addons if the following conditions are met:
-    //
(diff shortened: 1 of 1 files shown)


agjohnson:
I'm not sure exactly on on the exception handling issue, I'm guessing the exception isn't bubbling up through their evented interface -- this is common in async.

This might also be a place to use a stream interface:
https://developers.cloudflare.com/workers/runtime-apis/streams/

This is just a guess, and I don't know if it could be a drop in replacement for `transform()` (where I believe the exception originates), but it should :crossed_fingers: avoid consuming the whole page at once.

agjohnson:
Stream definitely seem like the way to avoid this exception in the first place. However, this is not quite the correct way to catch errors, as the exception emits as `response.body` is finally consumed -- this happens inside the `fetch` event handler.

I am currently trying to fix up this script and a number of bugs and that work
 will supersede this PR, I'll close this one.

PR #227, opened 2024-10-23
Clean up addons injection script and fix bugs in library

### LINKED ISSUE #225: Add tests for Addons injection
The working script would feel sturdier with tests, there's a lot happening in the script that we aren't verifying. I also found it a little hard to guarantee changes would work as expected without any test coverage.

Cloudflare has integration with Vite, which is slightly different than web-test-runner but the test cases do use similar syntax (Jest). We'd want to mock proxied request responses, here is an example to work from:

https://github.com/cloudflare/workers-sdk/blob/main/fixtures/vitest-pool-workers-examples/request-mocking/test/declarative.test.ts

agjohnson:
There is a lot here to modernize this library and fix a number of bugs: 

- Sets this up as a package so we can use normal package management to define dependencies.
- Adds a test framework to help catch bugs. I also found this necessary just to understand what the worker was doing and verify changes.
- Reworks the docker container and image so that wrangler can gracefully catch file changes -- it's not necessary to constantly restart wrangler now.
- Switch worker from event based to ESM layout -- don't use `fetch` event and instead use `fetch()` exported function
- Alter how we are returning `Response` instances, and when this happens.

----

- Fixes https://github.com/readthedocs/readthedocs.org/issues/11693
- Fixes https://github.com/readthedocs/addons/issues/410
- Refs https://github.com/readthedocs/readthedocs-ops/issues/1513
- Fixes #225 
- Does not address #226 yet


packages/addons-inject/index.js
agjohnson, line 1: GitHub diff is :-1: here

This is better, but not by much:

https://github.com/readthedocs/common/pull/227/files/ab4c393877c8c145c9688f8a5e30908518672327..c1c14e93884535acf283ac3095d134885bd97ae6

agjohnson:
I've tested this against projects with large response pages and the exception is actually caught now, and the response falls back to the origin response. The next change will be trying to catch this through streaming and avoiding buffering the entire request.

packages/addons-inject/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM node:18.15
+COPY package.json /usr/src/app/packages/addons-inject/
+COPY index.js /usr/src/app/packages/addons-inject/
+COPY wrangler.toml /usr/src/app/packages/addons-inject/
```
stsewd, line 4: Shouldn't this be volume?

packages/addons-inject/index.js
```diff
@@ -0,0 +1,309 @@
...
+
+  // Check to decide whether or not inject Addons library. We only do this for
+  // `text/html` content types.
+  if (contentType.includes("text/html")) {
+    // Remove old implementation of our flyout and inject the new addons if the following conditions are met:
+    //
+    // - header `X-RTD-Force-Addons` is present (user opted-in into new beta addons)
+    // - header `X-RTD-Hosting-Integrations` is not present (added automatically when using `build.commands`)
+    //
+    if (forceAddons === "true" && injectHostingIntegrations === "false") {
```
stsewd, line 163: From https://github.com/readthedocs/readthedocs.org/blob/3b947be7e3603f4fbac15b6af9d1d441fa41db1d/readthedocs/proxito/middleware.py#L273-L339, looks like these headers are enabled one at the time (they both can't be true).

```suggestion
    // Remove old implementation of our flyout and inject the new addons if the following conditions are met:
    //
    // - header `X-RTD-Force-Addons` is present (user opted-in into new beta addons)
    // - header `X-RTD-Hosting-Integrations` is not present (added automatically when using `build.commands`)
    //
    if (forceAddons === "true") {
```

docstring should be updated as well.

packages/addons-inject/index.js
```diff
@@ -0,0 +1,309 @@
...
+  // Log some debugging data
+  console.log(`ContentType: ${contentType}`);
+  console.log(`X-RTD-Force-Addons: ${forceAddons}`);
+  console.log(`X-RTD-Hosting-Integrations: ${injectHostingIntegrations}`);
+  console.log(`HTTP status: ${httpStatus}`);
+
+  // Debug mode for some test cases. This is just for triggering an exception
+  // from tests. There might be a way to conditionally enable this, but I had
+  // trouble getting Wrangler vars to work at all.
+  const throwError = headers.get("X-RTD-Throw-Error");
```
stsewd, line 145: maybe put "debug" somewhere in the header/variable? :thinking:

packages/addons-inject/index.js
```diff
@@ -0,0 +1,309 @@
...
+          // error being thrown during the request to origin. There are cases
+          // that are difficult to model here, like a large page size timing out
+          // the request.
+          element(element) {
+            if (throwError) {
+              throw new Error("Manually triggered error in transform");
+            }
+          },
+        })
+        .transform(await response.clone());
```
stsewd, line 206: Do we need to clone the response? Looks like the previous code, we didn't have to.

packages/addons-inject/index.js
```diff
@@ -0,0 +1,309 @@
...
+    }
+  }
+
+  // Inject Addons if the following conditions are met:
+  //
+  // - header `X-RTD-Hosting-Integrations` is present (added automatically when using `build.commands`)
+  // - header `X-RTD-Force-Addons` is not present (user opted-in into new beta addons)
+  //
+  if (forceAddons === "false" && injectHostingIntegrations === "true") {
```
stsewd, line 215: Same as above, only one of these can be true at a time.
```suggestion
  // Inject Addons if the following conditions are met:
  //
  // - header `X-RTD-Hosting-Integrations` is present (added automatically when using `build.commands`)
  // - header `X-RTD-Force-Addons` is not present (user opted-in into new beta addons)
  //
  if (injectHostingIntegrations === "true") {
```

packages/addons-inject/index.js
```diff
@@ -0,0 +1,309 @@
...
+    return new HTMLRewriter()
+      .on("head", new addPreloads())
+      .on(
+        "head",
+        new addMetaTags(projectSlug, versionSlug, resolverFilename, httpStatus),
+      )
+      .transform(await response.clone());
```
stsewd, line 222: same as above, do we really need to clone the response?

packages/addons-inject/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@readthedocs/addons-inject",
```
stsewd, line 2: we should take over this namespace if we haven't already :)

packages/addons-inject/index.js
ericholscher, line 206: Does cloning here use additional memory? Would be useful to k now why this is necessary instead of just modifying it in place?

packages/addons-inject/tests/wrangler.test.js
stsewd, line 1: tests :heart_eyes:

ericholscher (approved):
I don't love all the explicit test logic in the actual code, but seems it's necessary from the comments around it. Overall this looks similar to what we had, and the logic seems the same. I'm not really knowledgable enough about Workers or JS patterns to review that stuff, but the flow seems reasonable, with a lot more async 😨

packages/addons-inject/index.js
ericholscher: Looks like maybe this warning is the reason? 

https://developers.cloudflare.com/workers/runtime-apis/context/#passthroughonexception

packages/addons-inject/Dockerfile
agjohnson: It's mounted as a volume in our docker compose file, and this volume will override this path. This Dockerfile was previously was just installing a hardcoded version of wrangler, the copy operations are just for the initial build.

packages/addons-inject/index.js
agjohnson: You're probably right, but I don't want to change these conditionals yet. I'm a little confused by the usage of both these headers.

packages/addons-inject/index.js
agjohnson: I'll leave this be too, I don't want to prune too much logic yet. We can open an issue to track if needed.

packages/addons-inject/index.js
agjohnson: Noted here. https://github.com/readthedocs/common/pull/227/files#r1815664667

Cloning a response uses memory if the stream is left unread.

packages/addons-inject/index.js
```diff
@@ -0,0 +1,309 @@
...
+// "readthedocsDataParse" is the "<script>" that calls:
+//
+//   READTHEDOCS_DATA = JSON.parse(document.getElementById('READTHEDOCS_DATA').innerHTML);
+//
+const readthedocsDataParse = "script[id=READTHEDOCS_DATA]:first-of-type";
+const readthedocsData = "script[id=READTHEDOCS_DATA]";
+
+async function onFetch(request, env, context) {
+  // Avoid blank pages on exceptions
+  context.passThroughOnException();
```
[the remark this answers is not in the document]
agjohnson: There was no one fix that stopped this bug, which is why this is a larger rewrite. The problems previously contributing to this bug:

- Trying to reuse the original response without cloning
- No passthrough on exceptions. This call only helps some cases but not this bug
- We weren't evaluating the transformed body, so the exception was raised outside our code anyways (when workerd evaluated the response)

packages/addons-inject/package.json
agjohnson: Yeah might as well, we use this scope on all our packages.

packages/addons-inject/index.js
agjohnson: Added https://github.com/readthedocs/common/pull/227/commits/0699fa0c978f378af4a2818b9866614600198f2e

packages/addons-inject/tests/wrangler.test.js
```diff
@@ -0,0 +1,256 @@
...
+      .get("http://test-builds.devthedocs.org")
+      .intercept({ path: "/en/latest/" })
+      .reply(200, "<html></html>", {
+        headers: {
+          "Content-type": "text/html",
+          "X-RTD-Force-Addons": false,
+          "X-RTD-Hosting-Integrations": false,
+        },
+      });
+    let response = await SELF.fetch(
```
humitos, line 28: Is this `SELF.fetch()` the "magic" that executes our worker? I'm not sure to understand _where_ the CF worker is executed in these tests.

packages/addons-inject/tests/wrangler.test.js
```diff
@@ -0,0 +1,256 @@
...
+    fetchMock
+      .get("http://test-builds.devthedocs.org")
+      .intercept({ path: "/en/latest/" })
+      .reply(200, "<html><head></head><body></body></html>", {
+        headers: {
+          "Content-type": "text/html",
+          "X-RTD-Force-Addons": true,
+          "X-RTD-Hosting-Integrations": false,
+          "X-RTD-Project": "test-builds",
+          "X-RTD-Version": "latest",
```
humitos, line 67: We could also add the check for the status code here.

packages/addons-inject/tests/wrangler.test.js
```diff
@@ -0,0 +1,256 @@
...
+      `<div role="main"><div x-ref="outer"><div class="admonition warning" x-ref="inner"></div></div></div>`,
+      `<div role="main" x-ref="outer-furo"><div class="admonition warning" x-ref="inner-furo"></div></div>`,
+      `<div id="main-content"><div><div><article x-ref="outer-book"><div class="admonition warning" x-ref="inner-book"></div></article></div></div></div>`,
+      `<div class="rst-versions" x-ref="flyout"></div>`,
+    ];
+
+    // Ensure we're testing all of the removals
+    expect(scripts.length).toBe(AddonsConstants.removalScripts.length);
+    expect(links.length).toBe(AddonsConstants.removalLinks.length);
+    expect(elements.length).toBe(AddonsConstants.removalElements.length);
```
humitos, line 107: Great idea!

packages/addons-inject/tests/wrangler.test.js
```diff
@@ -0,0 +1,256 @@
...
+
+    // Make sure elements were removed, using extra attrs because the whole
+    // string is not removed
+    matcher.to.contain(`x-ref="outer"`);
+    matcher.to.not.contain(`x-ref="inner"`);
+    matcher.to.contain(`x-ref="outer-furo"`);
+    matcher.to.not.contain(`x-ref="inner-furo"`);
+    matcher.to.contain(`x-ref="outer-book"`);
+    matcher.to.not.contain(`x-ref="inner-book"`);
+    matcher.to.not.contain(`x-ref="flyout"`);
```
humitos, line 159: I'm not sure to understand this even with the context in the comment. Can you expand on why we are checking this extra attrs? 😄

packages/addons-inject/tests/wrangler.test.js
```diff
@@ -0,0 +1,256 @@
...
+          "X-RTD-Hosting-Integrations": false,
+        },
+      });
+    let response = await SELF.fetch(
+      "http://test-builds.devthedocs.org/en/latest/",
+    );
+    expect(response.status).toBe(200);
+    expect(await response.text()).toBe(
+      `<html><head>${AddonsConstants.scriptAddons}</head><body></body></html>`,
+    );
```
humitos, line 54: We can add checks here for the headers `X-RTD-Force-Addons` and `X-RTD-Hosting-Integrations` as well; so we know they are passing through as basic checks.

packages/addons-inject/tests/wrangler.test.js
```diff
@@ -0,0 +1,256 @@
...
+      137, 80, 78, 71, 13, 10, 26, 10, 0, 0, 0, 13, 73, 72, 68, 82,
+    ]);
+    fetchMock
+      .get("http://test-builds.devthedocs.org")
+      .intercept({ path: "/en/latest/_/static/images/test.png" })
+      .reply(200, binaryData, {
+        headers: {
+          "Content-type": "image/png",
+          "X-RTD-Force-Addons": true,
+          "X-RTD-Hosting-Integrations": false,
```
humitos, line 242: I didn't see any test for `X-RTD-Hosting-Integrations: true`, we should add some if we are using that chunk of code. We can open an issue for this if it's outside the scope of this PR, tho.

humitos (approved):
This looks really good to me. I'm happy we have tests 🧪 now and a pattern to keep adding more 💯 

It seems the main change here is the usage of the lower-level API `fetch` and `.passThroughOnException()` that allow us to catch errors on a `try/catch` when transforming the `clone()`d response. I don't fully understand the details behind this, but I think I get the big picture here.

I left some suggestions and questions that will allow me to understand a little more some missing pieces.

I'm trying to test this locally, but it seems the paths for some of the Docker files are wrong. I'm getting this error when running `inv docker.build`:

```
=> ERROR [wrangler 2/6] COPY package.json /usr/src/app/packages/addons-inject/                                                                                                        0.0s
 => ERROR [wrangler 3/6] COPY index.js /usr/src/app/packages/addons-inject/                                                                                                            0.0s
 => ERROR [wrangler 4/6] COPY wrangler.toml /usr/src/app/packages/addons-inject/                                                                                                       0.0s
------                                                                                                                                                                                      
 > [wrangler 2/6] COPY package.json /usr/src/app/packages/addons-inject/:                                                                                                                   
------                                                                                                                                                                                      
------                                                                                                                                                                                      
 > [wrangler 3/6] COPY index.js /usr/src/app/packages/addons-inject/:                                                                                                                       
------                                                                                                                                                                                      
------
 > [wrangler 4/6] COPY wrangler.toml /usr/src/app/packages/addons-inject/:
------
failed to solve: failed to compute cache key: failed to calculate checksum of ref UPQU:UMYX:4FAE:CRS7:GXRC:2DRM:5B5J:CRJO:MLEW:4WNJ:5M75:ZC7N::p64btx7uytedu4ry4kp8bwbzq: "/wrangler.toml": not found
```

I think we need to use relative paths using `readthedocs.org` as working path.

dockerfiles/docker-compose.yml
```diff
@@ -35,17 +35,15 @@ services:
   wrangler:
     build:
       context: ${PWD}
```
humitos, line 37: We need to pass the correct context here to be able to use the `COPY` commands in the Dockerfile:

```suggestion
      context: ${PWD}/common/packages/addons-inject/
```

humitos:
After building the images properly I'm getting this error:

```
wrangler_1     | node:internal/modules/cjs/loader:1078
wrangler_1     |   throw err;
wrangler_1     |   ^
wrangler_1     | 
wrangler_1     | Error: Cannot find module '/usr/src/app/packages/addons-inject/wrangler'
wrangler_1     |     at Module._resolveFilename (node:internal/modules/cjs/loader:1075:15)
wrangler_1     |     at Module._load (node:internal/modules/cjs/loader:920:27)
wrangler_1     |     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
wrangler_1     |     at node:internal/main/run_main_module:23:47 {
wrangler_1     |   code: 'MODULE_NOT_FOUND',
wrangler_1     |   requireStack: []
wrangler_1     | }
wrangler_1     | 
wrangler_1     | Node.js v18.15.0
wrangler_1 exited with code 1
```

packages/addons-inject/package.json
agjohnson: I started this, someone already claimed `readthedocs` organization. I reached out to support on this.

dockerfiles/docker-compose.yml
```diff
@@ -35,17 +35,15 @@ services:
...
+      - ${PWD}/common/packages/addons-inject/:/usr/src/app/packages/addons-inject/
     networks:
       readthedocs:
+    working_dir: /usr/src/app/packages/addons-inject/
     command: [
       "wrangler",
       "dev",
-      "/usr/src/app/docker/force-readthedocs-addons.js",
-      "--name=force-addons",
-      "--latest",
```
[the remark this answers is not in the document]
agjohnson: Auto reload does work, it actually works a bit better now with the directory mounted. Mounting just the file required restarting the wrangler container for me originally, because the file would disappear in the container on save. Wrangler detects the file update successfully now.

packages/addons-inject/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "format": "prettier --write ."
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/readthedocs/common.git"
+  },
+  "author": "Read the Docs, Inc",
+  "license": "MIT",
+  "dependencies": {
+    "wrangler": "^3.82.0"
```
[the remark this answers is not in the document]
agjohnson: Was there a reason behind using an old version?

packages/addons-inject/index.js
```diff
@@ -0,0 +1,319 @@
...
+    // necessary or wanted yet.
+    console.debug("Response body was already null, passing through.");
+    return response;
+  }
+
+  try {
+    const transformed = await transformResponse(response);
+    // Wait on the transformed body for errors to evaluate. Timeout errors and
+    // errors thrown during transform aren't actually raised until the body is
+    // read. If there was no return, we should return the original response.
```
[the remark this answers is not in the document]
agjohnson: The documentation and comments on `transformResponse` cover this, but if the response is not transformed `null` is returned instead of passing around the original response.

packages/addons-inject/index.js
```diff
@@ -0,0 +1,319 @@
...
+          // This is only used for testing purposes. This is used to mimic an
+          // error being thrown during the request to origin. There are cases
+          // that are difficult to model here, like a large page size timing out
+          // the request.
+          element(element) {
+            if (throwError) {
+              throw new Error("Manually triggered error in transform");
+            }
+          },
+        })
```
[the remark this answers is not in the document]
agjohnson: Good catch!

dockerfiles/docker-compose.yml
```diff
@@ -35,17 +35,15 @@ services:
...
-      dockerfile: ${PWD}/common/dockerfiles/Dockerfile.wrangler
+      dockerfile: ${PWD}/common/packages/addons-inject/Dockerfile
     volumes:
-      - ${PWD}/common/dockerfiles/force-readthedocs-addons.js:/usr/src/app/docker/force-readthedocs-addons.js
+      - ${PWD}/common/packages/addons-inject/:/usr/src/app/packages/addons-inject/
     networks:
       readthedocs:
+    working_dir: /usr/src/app/packages/addons-inject/
     command: [
       "wrangler",
```
humitos, line 45: We need this to be as follows to work properly when executing `inv docker.up`

```suggestion
      "node_modules/.bin/wrangler",
```

humitos:
> Just a note that **before merging this PR** we should fix the wrangler Docker container so it works with the pattern we already have and builds correctly.

I fixed this and left the solution as suggestions. I was able to build the Docker images and run it as usual with those changes.

packages/addons-inject/tests/wrangler.test.js
agjohnson: > Also, where we indicate that our CF worker lives in the index.js?
 
This is the `wrangler.toml` that dictates this.

> Is this SELF.fetch() the "magic" that executes our worker?

Yup, this executes `wrangler dev` pretty much. You can read more on this pattern here:

https://developers.cloudflare.com/workers/testing/integration-testing/

packages/addons-inject/package.json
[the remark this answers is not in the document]
humitos: I don't remember exactly. I know that we needed _at least_ this version because some of the previous one had an issue that we were hitting.

I suppose it's not important to upgrade. We have tests now and it's working fine, so 👍🏼

packages/addons-inject/tests/wrangler.test.js
agjohnson: When removing `<script>` elements, the change removes the entire element. From:

```
<head><script src="..."></script></head>
```

When removing HTML elements, we're only targeting the child elements for removal:

That is, a source:

```
<body><div><div><div class="rst-versions"></div></div></div></body>
```

After processing and removing the child flyout element, the parent `<div>` elements still exist:

```
<body><div><div></div></div></body>
```

The attributes help test for elements explicitly instead of guessing how each remaining `<div><div></div></div>` aligns with each source example.

packages/addons-inject/tests/wrangler.test.js
agjohnson: There already is one on line 73

packages/addons-inject/index.js
```diff
@@ -0,0 +1,319 @@
...
+    const transformed = await transformResponse(response);
+    // Wait on the transformed body for errors to evaluate. Timeout errors and
+    // errors thrown during transform aren't actually raised until the body is
+    // read. If there was no return, we should return the original response.
+    if (transformed !== undefined) {
+      return new Response(await transformed.arrayBuffer(), transformed);
+    }
+  } catch (error) {
+    console.error("Discarding error:", error);
+    console.debug("Returning original response to avoid blank page");
```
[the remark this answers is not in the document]
agjohnson: I did. Is there a specific concern here?

packages/addons-inject/index.js
agjohnson: Yup agreed.

I do think we should consolidate or at least rename these. It's not clear what the difference is between `forceAddons` and `injectHostingIntegrations`, the naming is confusing. I'm guessing these, and respective headers, should have been closer to `areAddonsEnabled` and `isUsingBuildCommands`.

packages/addons-inject/index.js
agjohnson: Also though, I'll note that this logic does feel inaccurate now. I am a bit confused by the variables/headers used here though. Is this saying that when the user has opted out of addons but is using build.commands, they still get addons?

If so, it seems like this logic won't matter. Almost all projects have addons enabled now, save for some that have asked to have addons disabled manually. So almost all projects should skip this logic, and if they set addons to be disabled, it feels weird to magically enable it again.

Shouldn't this be `forceAddons === "true"`?

packages/addons-inject/tests/wrangler.test.js
agjohnson: The issue here is that the logic we'd be testing doesn't seem to quite match what we're doing now. I'm a little confused by the variables/headers used here though.

I noted more here:
https://github.com/readthedocs/common/pull/227/files#r1819843063

I'll add a quick test, but I think that logic sounds wrong now.

agjohnson:
Review seems settled and the production test is looking much better:

image

I'm going to merge this and we can figure out the deploy next.

packages/addons-inject/index.js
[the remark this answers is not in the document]
humitos: Nope, just wondering if you were able to hit this try/catch in production to know that everything is working as expected -- since I had issues to handle the exceptions even when the code looked good.

packages/addons-inject/tests/wrangler.test.js
humitos: I mean the HTTP header that comunicates the status code, not the `response.status`.

I'm taking about the META HTML tag, 

```
<meta name="readthedocs-http-status" content="${this.httpStatus}" />
```

packages/addons-inject/index.js
humitos: > Is this saying that when the user has opted out of addons but is using build.commands, they still get addons?

I understand you are correct here. Projects using `build.commands` can't disable addons. It was built like that originally to start testing addons on builds we don't have any control.

> Almost all projects have addons enabled now, save for some that have asked to have addons disabled manually. So almost all projects should skip this logic, and if they set addons to be disabled, it feels weird to magically enable it again.

We should probably open an issue to talk more about this and probably remove this code now. I need to think more, tho, before being 100% sure it's like this 😄

PR #232, opened 2024-11-09
Install package globally instead

agjohnson:
I finally realized why I am getting different results here. It's because I was already doing local development on the package locally, so when I mounted my path, which included a `node_modules` path, everything kept working in the container.

If you do not have this path locally, when the docker compose container starts, your local path overlays the image's `node_modules` (and all of the installed packages there).

Instead, we can avoid this entirely by instead installing the package globally, outside the package path.

PR #235, opened 2024-11-15
Fix Wrangler Docker

ericholscher:
Using ENTRYPOINT broke a bunch of stuff on my side..


ericholscher:
@agjohnson I saw you changed this recently, but it blew up for me. I don't think we want to mess with ENTRYPOINT? We aren't using it anywhere else.

agjohnson:
Entrypoints should not be the issue here, I have a fix open here that I feel should be merged: https://github.com/readthedocs/common/pull/232

agjohnson (changes requested):
I'd rather not undo this and would merge #232, give that a try first.

PR #238, opened 2024-11-21
Follow current pattern for wrangler container

humitos:
I update the wrangler Docker configuration to follow the pattern we already have
for other containers. It mounts only the files that are needed to work and
nothing else.

This is done as an alterntive to #232
that should work for everybody.


ericholscher (approved):
This seems like a better approach to me :+1:

humitos:
> This does not solve the issue of wrangler not restarting for me 😞

I'm happy to pair with you to help on this, if you want.

humitos:
BTW, note that wrangler does not restart the process. It just reloads the file automatically. No message is print in the console, it does it automatically.

PR #240, opened 2024-11-27
Docker: standardize Docker compose configuration

humitos:
- Install webpack dependencies inside the Docker image
- Do not call `npm ci` for webpack at startup
- Mount addons volume under `checkouts/` as other containers
- Mount only required directories/files for ext-theme

### DIFF
--- a/dockerfiles/Dockerfile.webpack
+++ b/dockerfiles/Dockerfile.webpack
@@ -1,4 +1,4 @@
-FROM node:16
+FROM node:22.11
 COPY package-lock.json /usr/src/app/checkouts/ext-theme/
 COPY package.json /usr/src/app/checkouts/ext-theme/
 WORKDIR /usr/src/app/checkouts/ext-theme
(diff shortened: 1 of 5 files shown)


humitos:
ext-theme (webpack) starts pretty fast now. Basically, it only takes the time to start the webpack development server only 🚀

dockerfiles/Dockerfile.webpack
```diff
@@ -0,0 +1,5 @@
+FROM node:16
```
ericholscher, line 1: Might be worth noting why we're still in 16 here.

dockerfiles/docker-compose-webpack.yml
```diff
@@ -4,13 +4,16 @@
 services:
 
   webpack:
-    image: node:16
+    build:
+      context: ${PWD}/${RTDDEV_PATH_EXT_THEME:-../ext-theme}
+      dockerfile: ${PWD}/common/dockerfiles/Dockerfile.webpack
     working_dir: /usr/src/app/checkouts/ext-theme
```
ericholscher, line 10: Do we need to define this in the compose and dockerfile?

ericholscher (approved):

dockerfiles/Dockerfile.webpack
humitos: Ah, I may be confused with `website` maybe 🤔 . I updated this image to use 22.11 and it worked 💯

PR #241, opened 2024-11-27
Docker: auto-detect changes in dependencies and re-build

humitos:
Use the `watch` feature from Docker to auto-detect changes in
dependencies (NodeJS and Python) and rebuild the corresponding image.

If you change `readthedocs.org/requirements/docker.txt` you will see this message:

```
⦿ Rebuilding service "server" after changes were detected...
⦿ service "server" successfully built
```

If you change `ext-theme/package.json` you will see:

```
⦿ Rebuilding service "webpack" after changes were detected...
⦿ service "webpack" successfully built
```

Documentation: https://docs.docker.com/compose/how-tos/file-watch/


humitos:
I started exploring the idea of "watching the dependencies files and trigger an image rebuild automatically". This should work pretty fast most of the cases since it will use the cached layers when possible. @readthedocs/core let me know your thoughts.

humitos:
I played a little more with this and it seems that since the `server` built image is used by _other_ containers (web, celery, build, etc) Docker is not smart enough to restart those containers once the `server` image is re-built. It shows the following in the terminal:

```
web_1 exited with code 137
web_1 has been recreated
```

However, even if the `web` container was re-created, `docker ps` doesn't show it running 🤷🏼 . I tried adding `restart: always` to these services, but I got the same behavior.

With all of that, I can say that Docker Watch is useful only for our NodeJS containers for now. If we want to use it for our Python container, we will need to explore a little more why these services are not restarting properly.

ericholscher:
It seems like this only works when docker is actively running? It seems most of the time we need to rebuild the images because the requirements have changed in a PR, and we've just pulled it in? I'm not really sure this is something that happens frequently enough that we need to have a service for watching and rebuild images?

I think that having a fast rebuild with `uv`, and docker caching everything that isn't the python requirements mostly solves this for me without additional overhead.

dockerfiles/Dockerfile.webpack
```diff
@@ -2,4 +2,4 @@ FROM node:22.11
 COPY package-lock.json /usr/src/app/checkouts/ext-theme/
 COPY package.json /usr/src/app/checkouts/ext-theme/
 WORKDIR /usr/src/app/checkouts/ext-theme
-RUN npm ci
+RUN npm install
```
agjohnson, line 5: As long as we are checking in built assets to the `ext-theme` repository, this command should be `npm ci`.

`npm install` will upgrade previously installed packages without warning.

agjohnson:
I was going to point out the same thing about a running environment. Tuning the Dockerfile for caching would get us closer I feel.

Using `uv` could reduce our install time if the bottleneck isn't just network time. A smaller change would be splitting up our dependency installation steps into duplicate commands: one cached in the build, and one command that always runs. There are a few ways to accomplish cache busting, `COPY` is one, `ARG` is another common solution, or using two stage builds:

The discussion on this feature and workarounds is over a decade old, this shows `ARG`:

- https://github.com/moby/moby/issues/1996#issuecomment-185872769

humitos:
> It seems like this only works when docker is actively running?

Yes, this is dumb. There is a `--watch` attribute for `docker compose up`, but it doesn't check for changes at startup 🤷🏼 .

humitos:
I've opened a few PRs that use `uv` to install dependencies and they building time is a lot better now. I'd say the next steps to improve the build time would be using `ARG` as @agjohnson mentioned. This will be more useful on corporate since we download the dependencies file from GitHub there instead using it from the context because it's outside the current path (maybe a symlink works, tho)

PR #242, opened 2024-11-27
Development: run new dashboard by default

humitos:
- Make the new dashboard to run by default when running the development instance
- Always run the new dashboard using webpack
- Add `--legacy-dashboard` option to switch to the old/legacy dashboard


dockerfiles/Dockerfile.webpack
```diff
@@ -0,0 +1,5 @@
+FROM node:16
```
ericholscher, line 1: I thought we changed this to a more modern node? Not sure where that change was tho, seems like it would apply here?

dockerfiles/tasks.py
```diff
@@ -63,12 +61,9 @@ def up(c, search=True, init=False, reload=True, webpack=False, ext_theme=False,
 
     if search:
         cmd.append(f'-f {DOCKER_COMPOSE_SEARCH}')
-    if webpack:
+    if not legacy_dashboard:
         # This option implies the theme is enabled automatically
```
ericholscher, line 65:
```suggestion
```

ericholscher (commented):
Looks like this has some conflicts, I'm guessing with the changes I'm commenting on. Will wait for those to resolve.

humitos:
Updated 👍🏼

dockerfiles/tasks.py
```diff
@@ -39,16 +38,15 @@ def down(c, volumes=False):
...
-    'ext-theme': 'Enable new theme from ext-theme (default: False)',
+    'legacy-dashboard': 'Use the old/legacy dashboard (default: False)',
     'scale-build': 'Add additional build instances (default: 1)',
     'http-domain': 'Configure a production domain for HTTP traffic. Subdomains included, '
                    'example.dev implies *.examples.dev for proxito. Example: '
                    '"17b5-139-47-118-243.ngrok.io"',
     'log-level': 'Logging level for the Django application (default: INFO)',
     'django-debug': 'Sets the DEBUG Django setting (default: True)',
 })
-def up(c, search=True, init=False, reload=True, webpack=False, ext_theme=False, scale_build=1, http_domain="", django_debug=True, log_level='INFO'):
```
ericholscher, line 51: I think this will break with people's existing local dev workflows, so probably also worth a note in Slack to the team about this.

PR #243, opened 2024-11-28
Docker: reduce stop grace period to 1s

humitos:
There is no need to wait 10s for each container to stop. We aren't saving anything on disk on these containers, so we can kill them immediately.


humitos:
I don't know the exact reason, but I guess that Celery is has something to do here. Those containers running Celery are the ones that take the most to stop.

PR #246, opened 2024-12-05
Development: mount NPM dependencies files

humitos:
This is to allow installing dependencies from inside the container and perform the changes in the `package.json` files from the respository itself; so we can push these changes remotely after installing a new dependency.

Covers this use case
https://github.com/readthedocs/common/issues/245#issuecomment-2519778470

### DIFF
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -62,6 +62,8 @@ services:
       readthedocs:
     volumes:
       - ${PWD}/${RTDDEV_PATH_ADDONS:-../addons}/src/:/usr/src/app/checkouts/addons/src/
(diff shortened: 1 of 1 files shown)


agjohnson:
Note: this doesn't address what I'm describing in #245, but for that particular comment it still doesn't quite workaround the problem. It requires installing the package on my host (so I have the package in my host `node_modules/` and can run tests/etc) and also run the package installation command inside the container (so the image has the same dependencies).

Updating `package.json` from the container doesn't quite remove any steps, it still requires package installation on the host also. This is because we're not sharing `node_modules/` -- which is correct, we _should not_ share this between host and container.

humitos:
I'm merging this since this matches what we are doing with Python containers as well. We can continue exploring the solution proposed in #245.

PR #256, opened 2025-02-11
Upgrade Postgres to 16.3 to match prdouction

ericholscher:
This requires a couple changes:

* Before applying this, you can backup your data with ``inv docker.manage --backupdb shell``. Confirm it worked with `!ls`, which should show a `dump_*.sql` file
* If you keep the same data dir, postgres gets annoyed about the data dir being old, so this swaps a new one.
* After applying, you can run:

```
    inv docker.shell --container database --no-running

    # from another shell
    docker cp dump_*.sql community_database_1:/tmp/dump.sql

   # From inside container
    createdb -U docs_user docs_db
    psql -U docs_user docs_db < dump_*.sql
    psql -U docs_user ALTER USER docs_user WITH ENCRYPTED PASSWORD 'docs_pwd';
```

### DIFF
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -1,7 +1,7 @@
 volumes:
   build-user-builds:
   storage_data:
-  postgres_data:
+  postgres_data_16:
   postgres_backups_data:
 
 services:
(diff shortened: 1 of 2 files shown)


humitos (approved):
Cool! I migrated the data into the new pgdb and it worked great!

PR #267, opened 2025-04-10
Upgrade django-upgrade to latest version and Django 5.2

humitos:
Related https://github.com/readthedocs/readthedocs.org/issues/12079


humitos:
I'm merging it since we are already on Django 5.x

PR #273, opened 2025-06-19
Add Vale to RTD pre-commit

ericholscher:
Refs https://github.com/readthedocs/readthedocs.org/pull/12257

Downstrream:

* https://github.com/readthedocs/readthedocs-corporate/pull/2013
* https://github.com/readthedocs/readthedocs.org/pull/12258


pre-commit-config.yaml
```diff
@@ -1,3 +1,6 @@
+# It seems this repo is only used on .org?
```
humitos, line 1:
```suggestion
# It seems this file is only used on .org?
```

humitos: By the way, it's used in corporate as well https://github.com/readthedocs/readthedocs-corporate

humitos (approved):
Looks good 👍🏼

pre-commit-config.yaml
ericholscher: I assume this will break on corporate when merged, then? :/

humitos: Why do you think it will break? It may fail the linting, tho.

However, we don't update `common/` immediately, so we will need to update the common commit on corporate first to make this take effect.

humitos: Ah, yeah. You are right.

vale/RTD/sso.yaml
agjohnson, line 1: I updated this file in a PR, so we should reflect that here. But note that the filename is incorrect, it needs to be `.yml` for vale to find it.

vale/RTD/sso.yaml
```diff
@@ -0,0 +1,9 @@
+extends: substitution
+message: Use 'Single sign-on' instead of '%s'
```
agjohnson, line 2: Also note the message double substitution is important. Very silly this isn't just a default for the `message` attr.

```suggestion
message: Use '%s' instead of '%s'
```

PR #270, opened 2025-05-08
Fix docker.test

ericholscher:

PR #275, opened 2025-06-24
Restore previous ES option

stsewd:
This breaks ES, as -XX:UseSVE isn't recognized as a valid Java option.

Ref https://github.com/readthedocs/common/pull/270


ericholscher:
I think it's a Mac thing :(

PR #276, opened 2025-06-26
Fix bad pre-commit config

agjohnson:
We replaced all of `repos` key with the last update, this removed all other repo configurations.


ericholscher:
Bleh, my bad :(

PR #293, opened 2026-03-27
Make compilebuildtool build all tools by default

ericholscher:
## Summary
- Make `tool` and `version` arguments optional in `inv docker.compilebuildtool`
- When called without arguments, compiles all tools and all versions
- When called with just a tool (e.g. `inv docker.compilebuildtool python`), compiles all versions of that tool
- Prints progress (`[1/48] Compiling python 2.7...`) for visibility during long runs
- Previous behavior of specifying both tool and version still works exactly as before

Per @humitos in readthedocs/readthedocs.org#11235: the command should build everything by default.

## Test plan
- [ ] `inv docker.compilebuildtool` (no args) compiles all tools and versions
- [ ] `inv docker.compilebuildtool python` compiles all Python versions
- [ ] `inv docker.compilebuildtool python 3.12` compiles just Python 3.12 (existing behavior)
- [ ] `inv docker.compilebuildtool --version 3.12` without tool prints error message

Made by AI

### DIFF
--- a/dockerfiles/tasks.py
+++ b/dockerfiles/tasks.py
@@ -189,12 +189,17 @@ def test(c, arguments='', running=True):
         c.run(f'{DOCKER_COMPOSE_COMMAND} run -e GITHUB_USER=$GITHUB_USER -e GITHUB_TOKEN=$GITHUB_TOKEN --rm --no-deps web uv run tox {arguments}', pty=True)
 
 @task(help={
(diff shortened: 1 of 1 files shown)


humitos:
Preliminar review, looks good.

humitos:
Seems you started from an old commit since there are a few commits that are mine 😄

ericholscher:
I think I ended up copying them from S3 with rclone, which is probably a better approach.

PR #302, opened 2026-05-14
Serve docs properly with El Proxito

humitos:
I hit an issue during the offsite that all my local docs under development return 404 and Claude solved the issue by downgrading wrangler.


ericholscher (approved):

ericholscher:
Janky comment syntax :D

stsewd:
We probably want an issue to track the upgrade, we want to keep up to date with the latest version of wrangler for better test the integration with CF.

PR #304, opened 2026-05-22
Fix django-debug-toolbar by reducing worker count

agjohnson:
Debug toolbar has been pretty useless for a long time now, all pages in
the toolbar mostly show failures and only work a small portion of the
time. Reducing the worker count resolves this behavior and allows all
pages to function.

There might be a longer term fix here, the problem is likely the workers
not sharing some state that debug toolbar is expecting.

### DIFF
--- a/dockerfiles/entrypoints/web.sh
+++ b/dockerfiles/entrypoints/web.sh
@@ -13,7 +13,7 @@ then
     uv run python3 manage.py loaddata test_data
 fi
 
-CMD="uv run gunicorn readthedocs.wsgi:application -w 3 -b 0.0.0.0:8000 --max-requests=10000 --timeout=0"
+CMD="uv run gunicorn readthedocs.wsgi:application -w 1 -b 0.0.0.0:8000 --max-requests=10000 --timeout=0"
 
 if [ -n "${DOCKER_NO_RELOAD}" ]; then
   echo "Running process with no reload"


stsewd (approved):
I think it was set to multiple workers to have something more similar to production to catch errors more early. We could have a debug mode that enables DDT and sets the number of workers to 1 too, but I'm fine with this too.

agjohnson:
Yeah, though if we special case it, we likely won't use it. There might be a proper fix at the debug toolbar eventually too. But I'm fine getting back to that later and fixing the toolbar for now.

I think the toolbar should be quite important for development on the templates. It seemed like a struggle to debug templates without the toolbar, I'd like to make sure we're all using a workflow that actually works. The toolbar is the best way to inspect what templates are used and the context data to them, it's just unfortunate that it doesn't work with multiple workers currently.

dockerfiles/entrypoints/web.sh
```diff
@@ -13,7 +13,7 @@ then
     uv run python3 manage.py loaddata test_data
 fi
 
-CMD="uv run gunicorn readthedocs.wsgi:application -w 3 -b 0.0.0.0:8000 --max-requests=10000 --timeout=0"
+CMD="uv run gunicorn readthedocs.wsgi:application -w 1 -b 0.0.0.0:8000 --max-requests=10000 --timeout=0"
```
humitos, line 16: It would be good to add a comment here explaining why we are using only 1 worker.

humitos (approved):
Looks good after adding a comment in that file so we know why we are using only one worker.

agjohnson:
Took another stab at resolving this, using cache storage seems promising. Trying out a PR at:

- https://github.com/readthedocs/readthedocs.org/pull/13087

Seems to fix the issue for me.

PR #307, opened 2026-06-02
Agents: add PR description guidance on rationale over restatement

ericholscher:
`AGENTS.md` is copied into every production repo, so its PR section shapes every agent-written PR. The guidance was loose enough to produce bloated descriptions that just restate the diff — and, twice in this PR's own history, duplicate "generated by AI" footers. This sharpens that section to ask for the *why* over a restatement of the *what*, and fixes the footer rule so a second one never gets appended. A few stale lines in the surrounding text are corrected in passing.

`AGENTS.md` is the canonical copy and there's no separate `CLAUDE.md` here, so this is the only place to change.


AGENTS.md
```diff
@@ -96,7 +96,6 @@ before they cause `IntegrityError` failures at deploy time.
 - Use Django conventions and best practices
 - Use type hints for function signatures
 - Write clear, concise docstrings for public functions and classes
-- Run linters and formatters using `tox -e pre-commit` before committing code
```
ericholscher, line 99: There were multiple mentions of pre-commit so removed this to make it explicit.

AGENTS.md
```diff
@@ -11,7 +11,7 @@ Read the Docs is a documentation hosting platform that builds and hosts document
 It supports multiple documentation tools (Sphinx, MkDocs, etc.) and automatically builds documentation from Git repositories.
 
 **Technology Stack:**
-- Python 3.x
+- Python 3.12
```
humitos: Interesting. We upgrade to Ubuntu 26.04 which defaults to 3.14 -- However, we are using `uv` now and we haven't upgraded it there: https://github.com/readthedocs/readthedocs-ops/blob/main/salt/base/readthedocs/common.sls#L45

We are using 3.14 on development now, so we should probably upgrade production as well.

humitos (approved):

AGENTS.md
ericholscher:
```suggestion
- Python 3.14
```

humitos: Actually, I have a PR opened at https://github.com/readthedocs/readthedocs.org/pull/13027/ to use Python 3.14 locally and test everything works as expected 👍🏼

PR #308, opened 2026-06-02
Agents: trim boilerplate that agents already follow

ericholscher:
HS: I asked the agent how to improve the Agent file, and this is what it came up with. And they can make sense, we should only be stating things that are unique to our codebase. It knows to follow the existing style.

`AGENTS.md` tells agents to cut filler, then spends whole sections on advice any competent agent already follows — PEP 8, "be aware of OWASP", "write descriptive test names." That boilerplate buries the rules that are genuinely RTD-specific and easy to get wrong (the `tox` recipes, the migration check, the marketing-names rule). Removing it so the guidance that remains is worth reading.

If any of the dropped sections was load-bearing for a reason I'm missing, say so and I'll restore it — ideally replaced with a concrete, RTD-specific rule rather than a placeholder.

Stacked on #307 to avoid conflicts on the same file; rebase onto `main` once that lands.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -55,12 +55,8 @@ It supports multiple documentation tools (Sphinx, MkDocs, etc.) and automaticall
 
 ## Testing
 
-- Use `pytest` as the testing framework
-- Use the `assert foo == bar` style for assertions
-- Write succinct, focused tests that test one thing at a time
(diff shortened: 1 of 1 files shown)


humitos (approved):

ericholscher:
@copilot resolve the merge conflicts in this pull request

PR #310, opened 2026-06-02
Several fixes for docker builds and Node.js

agjohnson:
- Use Node.js 24.16.0 for images
- Drop comment in package.json, it throws an exception and blocks building

### DIFF
--- a/packages/addons-inject/package.json
+++ b/packages/addons-inject/package.json
@@ -18,7 +18,6 @@
   "author": "Read the Docs, Inc",
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


packages/addons-inject/package.json
```diff
@@ -18,7 +18,6 @@
   "author": "Read the Docs, Inc",
   "license": "MIT",
   "dependencies": {
-    "//": "wrangler is pinned: 4.20+ rewrites inbound Host and X-Forwarded-Host to the value of `dev --host`, which breaks subdomain-based project resolution in El Proxito for local development",
```
humitos, line 21: :(

humitos (approved):

PR #311, opened 2026-06-15
Builds: isolated builders (development environment)

humitos:
Required by: https://github.com/readthedocs/readthedocs.org/pull/13133


dockerfiles/docker-compose.yml
```diff
@@ -214,6 +227,9 @@ services:
       - ${PWD}/common/dockerfiles/nodemon.json:/usr/src/app/checkouts/nodemon.json
       - ${PWD}/${RTDDEV_PATH_EXT:-../readthedocs-ext}:/usr/src/app/checkouts/readthedocs-ext
       - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../ext-theme}:/usr/src/app/checkouts/ext-theme
+
+      # Docker in Docker: Fargate (call `docker run` from inside `web` queue)
+      - /var/run/docker.sock:/var/run/docker.sock
```
humitos:
```suggestion
```

agjohnson: This still references Fargate, is this old?

dockerfiles/docker-compose.yml
```diff
@@ -362,6 +380,81 @@ services:
...
+    stop_grace_period: 1s
+    build:
+      context: ${PWD}
+      dockerfile: ${PWD}/common/dockerfiles/Dockerfile.isolated-builder
+    volumes:
+      # Host's readthedocs-builder checkout. The entrypoint sees the
+      # mount as populated and skips ``git clone`` so you can iterate
+      # on the runner / worker code by editing on the host and
+      # restarting the container.
+      - ${PWD}/${RTDDEV_PATH_BUILDER:-../readthedocs-builder}:/opt/readthedocs-builder
```
agjohnson: This seems not super consistent, why the different path in `/opt/` here? We're using `/usr/src/app/` for the app, why not `/usr/src/builder/`?

dockerfiles/docker-compose.yml
```diff
@@ -362,6 +380,81 @@ services:
...
+      # this file). The entrypoint populates both via ``uv sync
+      # --python 3.14 --python-preference=only-managed``; the worker
+      # then bind-mounts them into each build container by the volume
+      # names ``community_runner-venv`` / ``community_uv-python``
+      # (passed through RTD_HOST_*_PATH below).
+      - runner-venv:/opt/runner-venv
+      - uv-python:/opt/uv-python
+
+      # The entrypoint script itself, bind-mounted so edits don't
+      # need a rebuild (matches how celery.sh / build.sh wire up).
```
agjohnson: How is this different from the normal pattern used in other services in this file?

Speaking generally on what I assume llm generated code, I think this is a case of overly verbose commenting from the llm. I'm noticing similar in a lot of output. Having a lot of comments to wade through actually makes it more confusing and harder work to process. We don't note these things anywhere else in the file, so I need to read through all these comments to figure out what is actually important and different from the rest of the file.

dockerfiles/docker-compose.yml
```diff
@@ -362,6 +380,81 @@ services:
...
+      # to the same checkout the bind-mount above points at.
+      - RTD_HOST_BUILDER_PATH=${PWD}/${RTDDEV_PATH_BUILDER:-../readthedocs-builder}
+      # Runner venv + managed Python install: host-daemon-recognized
+      # names of the named volumes defined at the top of this file.
+      # Compose prefixes volume names with the project (``community``
+      # for the default community-edition project). The worker passes
+      # these as ``-v <name>:<path>:ro`` and the daemon resolves them
+      # as named-volume mounts.
+      - RTD_HOST_RUNNER_VENV_PATH=community_runner-venv
+      - RTD_HOST_PYTHON_INSTALL_PATH=community_uv-python
```
agjohnson: The comment here isn't super clear. If these are volumes, then why are they referred to as paths? Shouldn't this be:


```suggestion
      # Runner venv + managed Python install: host-daemon-recognized
      # names of the named volumes defined at the top of this file.
      # Compose prefixes volume names with the project (``community``
      # for the default community-edition project). The worker passes
      # these as ``-v <name>:<path>:ro`` and the daemon resolves them
      # as named-volume mounts.
      - RTD_HOST_RUNNER_VENV_VOLUME=community_runner-venv
      - RTD_HOST_PYTHON_INSTALL_VOLUME=community_uv-python
```

humitos: Both work. Either a path on the host or a volume name.

We are mounting it as `--volume {PATH, VOLUME}:/opt/uv-python`. We are using a PATH in production and VOLUME on development.

We can probably use a PATH in development as well to make it more consistent and avoid this confusion.

humitos: Done in  323e113

dockerfiles/docker-compose.yml
humitos: Yes, I marked this to be deleted but hasn't deleted yet 👍🏼

dockerfiles/docker-compose.yml
humitos: I moved this to `/usr/src/builder` in 0762a41

dockerfiles/docker-compose.yml
humitos: There is no difference. We mount the `isolated-builder.sh` from the repository into the Docker container to be able to modify it (develop) without rebuilding the image/restarting the container.

Good point about the comments. I left them all for now because the were a lot of things changing from one pattern to another during the POC: Fargate, ECS, the Celery approach, etc. I thought it would be easier to follow while discussing about this.

I removed the ones that are not useful because they match what we are already doing in other services and keep the ones that are relevant to the new pattern in [`37cb869`](https://github.com/readthedocs/common/pull/311/changes/37cb86913e45709db53007c2ef87aa916d8a3911)

dockerfiles/docker-compose.yml
humitos: Removed in 9df47aa

dockerfiles/docker-compose.yml
agjohnson: Yeah that makes sense, commenting here around important changes for the isolated builders is still useful. It was strictly the extra comments that confused me

dockerfiles/docker-compose.yml
agjohnson: Ah gotcha yeah that makes sense it tries to match prod. The change looks good :+1:

dockerfiles/nodemon.json
```diff
@@ -2,7 +2,7 @@
     "verbose": false,
     "delay": 2000,
     "ext": "py",
-    "watch": ["readthedocs", "readthedocsinc", "../readthedocs.org/readthedocs", "../readthedocs-ext/readthedocsext"],
+    "watch": ["readthedocs", "readthedocsinc", "../readthedocs.org/readthedocs", "../readthedocs-ext/readthedocsext", "/opt/readthedocs-builder"],
```
agjohnson: Guessing this might need updated too:


```suggestion
    "watch": ["readthedocs", "readthedocsinc", "../readthedocs.org/readthedocs", "../readthedocs-ext/readthedocsext", "/usr/src/builder"],
```

agjohnson (approved):

dockerfiles/nodemon.json
humitos: Yeah, thanks.

PR #316, opened 2026-07-16
Extra changes for isolated builders on local development

humitos:
- ~**Mount rclone inside the container**: this is required because we are uploading the artifacts from inside the builder now. The `readthedocs/build:<os>` image doesn't have `rclone` installed. Instead of rebuilding all those images, the easiest solution is to install rclone in the image spinning up the Docker container and mount the binary there. *NOTE: we can find another way to do this later, but I want to have a solution in place for now. This is easiest one that I found.*~
- **Install rclone** inside the Docker image
- **Mount Docker socket**: this is required to allow the container to kill Docker containers
- **Pass extra environment variables**
- **Install `ssh`**: to be able to clone private repositories

### DIFF
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -128,6 +128,9 @@ services:
       - ${PWD}/common/dockerfiles/nodemon.json:/usr/src/app/checkouts/nodemon.json
       - ${PWD}/${RTDDEV_PATH_EXT:-../readthedocs-ext}:/usr/src/app/checkouts/readthedocs-ext
       - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../ext-theme}:/usr/src/app/checkouts/ext-theme
+
+      # Docker in Docker: cancel running builds.
+      - /var/run/docker.sock:/var/run/docker.sock
     links:
       - storage
       - database
@@ -215,7 +218,7 @@ services:
       - ${PWD}/${RTDDEV_PATH_EXT:-../readthedocs-ext}:/usr/src/app/checkouts/readthedocs-ext
       - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../ext-theme}:/usr/src/app/checkouts/ext-theme
 
-      # Docker in Docker
+      # Docker in Docker: Fargate (call `docker run` from inside `web` queue)
       - /var/run/docker.sock:/var/run/docker.sock
     links:
       - storage
@@ -331,7 +334,7 @@ services:
       # Because of this, we need to use a shared volume between them
       - build-user-builds:/usr/src/app/checkouts/readthedocs.org/user_builds
 
-      # Docker in Docker
+      # Docker in Docker: old implementation (`build-default` queue)
       - /var/run/docker.sock:/var/run/docker.sock
     links:
       - web
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -403,8 +403,10 @@ services:
       # image). Bind-mounted onto PATH inside each build container.
       - RTD_HOST_RCLONE_PATH=${PWD}/${RTDDEV_PATH_BUILDER:-../readthedocs-builder}/rclone/rclone
       # Have build containers join the compose network so they can
-      # reach the API at ``http://web`` via service-name DNS.
-      - RTD_BUILD_NETWORK=community_readthedocs
+      # reach the API at ``http://web`` via service-name DNS. Derived from the
+      # compose project so it's correct for both community (``community``) and
+      # commercial (``commercial``) — a hardcoded value breaks the other env.
+      - RTD_BUILD_NETWORK=${COMPOSE_PROJECT_NAME}_readthedocs
     stdin_open: true
     tty: true
     networks:
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -214,6 +214,9 @@ services:
       - ${PWD}/common/dockerfiles/nodemon.json:/usr/src/app/checkouts/nodemon.json
       - ${PWD}/${RTDDEV_PATH_EXT:-../readthedocs-ext}:/usr/src/app/checkouts/readthedocs-ext
       - ${PWD}/${RTDDEV_PATH_EXT_THEME:-../ext-theme}:/usr/src/app/checkouts/ext-theme
+
+      # Docker in Docker
+      - /var/run/docker.sock:/var/run/docker.sock
     links:
       - storage
       - database
@@ -256,6 +259,7 @@ services:
       - RTD_S3_STATIC_STORAGE_BUCKET
       - RTD_AWS_S3_REGION_NAME
       - RTD_OPENAI_API_KEY
+      - RTD_BUILDER_TOKEN
     stdin_open: true
     tty: true
     networks:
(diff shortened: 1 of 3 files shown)


humitos:
@stsewd this is the only PR that's missing to merge to have the new builder fully working locally. With that, running `inv docker.up` + adding a project to `ISOLATED_BUILDER` feature should run those project's builds in the isolated-builder container.

dockerfiles/entrypoints/isolated-builder.sh
```diff
@@ -41,6 +42,24 @@ else
...
+#     The runner syncs artifacts to storage with rclone from inside the
+#     build container, and the worker bind-mounts the binary in rather
+#     than each build downloading its own. In production Packer bakes
+#     rclone into the AMI, so the worker mounts a real host path.
+#
+#     Here the worker runs in THIS container, but ``docker run -v`` is
+#     resolved by the HOST daemon (docker-out-of-docker via the mounted
+#     socket) — so /usr/local/bin/rclone from this image is invisible to
+#     it. Copying into $RCLONE_DIR (a host bind-mount, see compose) puts
+#     the binary on the host filesystem, same trick as $RUNNER_VENV.
```
agjohnson: In general, I think we need to tune the pattern that the llms use to generate comments around code or do more editing. I keep seeing this level of commenting generated and it makes me have to read through a lot of not super helpful comments -- and so I'm going to miss what _is_ important if it's there. In this case, explaining docker and bind mounts is noisy information.

```suggestion
# This hack mounts the host rclone binary into the container so that it
# doesn't need to be installed into the container each build.
```

dockerfiles/entrypoints/isolated-builder.sh
```diff
@@ -41,6 +42,24 @@ else
...
+if [ ! -x "$RCLONE_DIR/rclone" ]; then
+    echo "[isolated-builder] Copying rclone to $RCLONE_DIR (host-visible for build containers) ..."
+    install -m 0755 /usr/local/bin/rclone "$RCLONE_DIR/rclone"
```
agjohnson: What does "host-visible" mean?

I didn't realize until looking over the compose file that this is not the host system that rclone comes from, but from the builder container. If so, mentioning the host here at all is confusing.

humitos: This is confusing due to docker-in-docker pattern we use in our Docker Compose: basically, a Docker container that spins up _another_ container.

This is what's happening here:
1. Docker Compose mounts a rclone volume
2. This `isolated-builder` containers copies the `rclone` binary _from inside the container_ into the host (rclone volume) at boot time
3. Isolated builder container spins up a new container (Ubuntu 26.04) mounting the `rclone` binary just copied to the host (rclone volume) into the the new Ubuntu 26.04 container

This is required because when spining up a new container from inside a container means that you are using Docker socket from the host; and the `rclone` binary from inside the container is not visible by the host --so we need to use a `rclone` binary that lives in the host. That `rclone` binary is the one that we just copied out from the container into the host.

This is complex; I know. I didn't find a better way to do it yet but I didn't want to spend too much time on this for now since it's just an implementation detail.

I noted that we should find a different/better/cleaner solution for this. Maybe we can just mount the original `rclone` from the host and avoid this dance. However, in that case, we will be making the development environment to depend on having `rsync` installed in the host --which is something I wanted to avoid. I prefer if we don't have external dependencies, and also we will need to deal with different versions, etc.

Am I clear? 😬

agjohnson: Yeah that matches what I gathered from the code here, that all makes sense.

But yeah I think this pattern could be simplified. I don't fully understand the need for this pattern at least in development. If this is simply that rclone isn't installed in the build image, we can fix this with installing rclone in the build image, no?

Is the time difference really that bad in local development that we can't just keep this way simpler and install rclone on build start in development? It's certainly not efficient but it's easy to understand.

humitos: > If this is simply that rclone isn't installed in the build image, we can fix this with installing rclone in the build image, no?

Unfortunately, no. We need `rclone` installed in the host because it's the one that spins the `readthedocs/build:ubuntu-*` container and needs to "see" the `rclone` binary to be able to mount it there. The host can't mount a binary that's inside a container.

> Is the time difference really that bad in local development that we can't just keep this way simpler and install rclone on build start in development? It's certainly not efficient but it's easy to understand.

Same answer as before 😄 -- we need the `rclone` binary visible at host level, not inside the container.

I mean, we need the `rclone` binary inside the `readthedocs/build:` images ideally --but we won't modify the old images just because of this to avoid potential breakages here. Since the host is the one that spins up `readthedocs/build:` containers, the host needs a `rclone` binary to mount on.

---

The "easy" way to solve this problem, is to require `rclone` to be installed in the host (your OS) for Read the Docs to work. We currently require Git, pip and Docker. We could add `rclone` to that list to avoid this dance 👯 [^1]

That said, I'm not seeing a clear simpler solution that the one implemented yet. That's unfortunate, tho.

[^1]: by the way, it seems it breaks on MacOS tho, since it runs inside a VM there. So, having the `rclone` at OS level won't work. I think we still have Eric using Mac, right?

humitos:
I updated this PR with the latest updates to use `docker exec` instead as we talked in Slack.

dockerfiles/entrypoints/isolated-builder.sh
agjohnson: > Unfortunately, no. We need rclone installed in the host because it's the one that spins the

This is a bit confusing, I think there is some terms mixed up here and in the comments through the files. I think you might be referring to the task worker container as 'host' here? The worker container isn't a host of anything, the build container is not running from the worker container. The worker container simply has control of the host's Docker daemon and can spawn the build container on the host.

So the task worker container needed rclone installed in order to mount it into build containers to avoid us altering the build image to install rclone? It still feels like the answer here is to just update our build image.

It looks like this design is probably gone away with moving back to docker exec execution from the task worker container though. If so there isn't much to change here.

humitos: > It looks like this design is probably gone away with moving back to docker exec execution from the task worker container though. If so there isn't much to change here.

Yes. I'm installing `rclone` in the Docker container now and all this dance is not required anymore: a4b20500ee48b67942db46bb8d39788e8fe99707

Actually, using `docker exec` simplified a few things that I wasn't expecting. I thought that it would make things more complex, but in fact it was the opposite.

dockerfiles/docker-compose.yml
```diff
@@ -256,7 +257,6 @@ services:
       - RTD_S3_STATIC_STORAGE_BUCKET
       - RTD_AWS_S3_REGION_NAME
       - RTD_OPENAI_API_KEY
-      - RTD_BUILDER_TOKEN
```
stsewd, line 259: Looks like we no longer pass this env var, but the script still checks for it.

stsewd (commented):
What's the reason we need to share these directories with the host?

https://github.com/readthedocs/common/blob/a4b20500ee48b67942db46bb8d39788e8fe99707/dockerfiles/docker-compose.yml#L374-L375

agjohnson (approved):
This is easier to follow with the changes to use docker exec 👍

dockerfiles/docker-compose.yml
humitos: Good catch. We don't really need this env var locally because we require to have the `readthedocs-builder` cloned. We can remove it from the script as well, I guess.

humitos:
> What's the reason we need to share these directories with the host?

These are leftovers from before. I will remove them.

PR #323, opened 2026-09-01
Vale: demote CustomSpelling from error to warning

ericholscher:
Unknown words keep blocking docs PRs — every new tool page trips on legitimate product names (Zensical, Astro, goss) and the fix requires editing the shared dictionary here, in a different repo than the failing PR.

Demoting the rule to a warning keeps misspellings visible in local runs and verbose CI output without failing builds over names that aren't in the dictionary. The regex word list stays in place, so anything already accepted continues to produce no noise at all.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01ENQry9mujpyas4QWUAn7sv


humitos:
I don't think we should block PRs from merging if the CI is failing for this small issues. We can merge them and fix the issue on this repository after that 👍🏼 

Lowering this down to warning, will not communicate us there is a linting issue and those will just accumulate over time since we won't even notice there is an issue.

I don't have strong feelings here, but just wanted to mention the downside of using warning here.

ericholscher:
Yea... we could break it out into a different failing non-required task in CI perhaps, instead?

humitos:
Yeah 👍🏼 -- that would be ideal. We will see the red X on GitHub but it won't block the merge 💯

PR #330, opened 2026-09-15
RustFS: pin to an older version that doesn't break s3s multipart uploads

humitos:
When working with the upload CLI, newer versions of RustFS break s3s multipart uploads, so we pin to an older version.

All the uploads were failing when working locally.

### Migration steps locally

There is some changes needed here due to permissions being wrong between different versions:

1. Update `common` and spin up docker

   ```bash
   git submodule update --remote common   # or pull the branch that has the pin
   inv docker.up
   ```

2. Backup old bucket metadata just in case

   ```bash
   docker exec community-storage-1 sh -c '
   cd /data/.rustfs.sys/buckets
   mkdir -p /data/.rustfs.sys/buckets-metadata-backup-alpha83
   for b in static media build-tools builds usercontent build-uploads; do
     mkdir -p /data/.rustfs.sys/buckets-metadata-backup-alpha83/$b
     mv $b/.metadata.bin /data/.rustfs.sys/buckets-metadata-backup-alpha83/$b/
   done'
   ```

3. Restart storage so it starts with default bucket metadata:

   ```bash
   docker restart community-storage-1
   ```

4. Re-apply the public-read policies with the normal setup script:

   ```bash
   docker exec -e INIT=1 -e RUSTFS_ROOT_USER=admin -e RUSTFS_ROOT_PASSWORD=password \
     community-web-1 sh /usr/src/app/docker/createbuckets.sh
   ```


stsewd:
Not sure we should downgrade. Don't we have a way to increase the limit? Does this happen with curl or just requests or just with big files? There may be some other options in the http client that could prevent the error from happening.

humitos:
> Don't we have a way to increase the limit?

I didn't find one.

> Does this happen with curl or just requests or just with big files?

This happens randomly with cURL and/or our own CLI and depends on how the TCP package is built before being sent from what I can tell. I found this by talking to AI, and, honestly, I didn't understand 100% of the issue because it's too low-level to me and I thought it didn't worth the time if there was a quick fix for now.

Let me tell the AI to write a more detailed document that I can share here in case you want to dig more into this. I don't think it worth our time, tho. The real solution is opening an issue on the `s3s` repository so they can fix it.

Here is the document the AI wrote explaining what we found: [RUSTFS-S3S-MULTIPART-BUG.md](https://github.com/user-attachments/files/32280841/RUSTFS-S3S-MULTIPART-BUG.md)

